PR #3, opened 2014-11-04
Support datadog extensions to statsd datagrams

pjjw:
Datadog's format extensions allow attaching labels and tags to statsd
metrics, and are documented at
http://docs.datadoghq.com/guides/dogstatsd/#datagram-format

Additionally changes compare method in tests to reflect.DeepEqual, as
the tag maps don't sort deterministically, corrects an inversion in
displaying test failure cases, and treats an invalid sample value as 1
instead of throwing out the sample.

### DIFF
--- a/bridge.go
+++ b/bridge.go
@@ -143,31 +143,38 @@ func (c *SummaryContainer) Get(metricName string, labels prometheus.Labels) prom
 type Event interface {
 	MetricName() string
 	Value() float64
+	Labels() map[string]string
 }
 
 type CounterEvent struct {
 	metricName string
 	value      float64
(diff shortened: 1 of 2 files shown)


prepor:
No chance to merge?

pjjw:
IIRC this code doesn't actually work except in tests.

sdurrheimer:
:+1: This feature would be a nice addition.

juliusv:
If this is wanted by people, someone who actually uses Datadog needs to contribute a working PR (and test it, etc.)...

grobie:
Fixed in #40

PR #7, opened 2015-02-11
Cosmetics

bracki:
* Use argument passed to function instead of global variable
* Use `fatalf`

@juliusv

### DIFF
--- a/main.go
+++ b/main.go
@@ -30,9 +30,9 @@ func serveHTTP() {
 }
 
 func udpAddrFromString(addr string) *net.UDPAddr {
-	host, portStr, err := net.SplitHostPort(*statsdListenAddress)
+	host, portStr, err := net.SplitHostPort(addr)
 	if err != nil {
-		log.Fatal("Bad StatsD listening address", *statsdListenAddress)
+		log.Fatal("Bad StatsD listening address", addr)
 	}
 
 	if host == "" {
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -30,9 +30,9 @@ func serveHTTP() {
 }
 
 func udpAddrFromString(addr string) *net.UDPAddr {
-	host, portStr, err := net.SplitHostPort(*statsdListenAddress)
+	host, portStr, err := net.SplitHostPort(addr)
```
juliusv: Oh, this was changed to `statsdListenAddress` in https://github.com/prometheus/statsd_bridge/pull/6/files. May need updating of your PR here?

main.go
```diff
@@ -45,7 +45,7 @@ func udpAddrFromString(addr string) *net.UDPAddr {
 
 	port, err := strconv.Atoi(portStr)
 	if err != nil || port < 0 || port > 65535 {
-		log.Fatal("Bad port %s: %s", portStr, err)
+		log.Fatalf("Bad port %s: %s", portStr, err)
```
juliusv: Thanks for catching that!

main.go
juliusv: ZOMG. Now I see what you did. All good.

juliusv:
Thanks, merging!

PR #8, opened 2015-02-18
Fix license headers to match LICENSE file.

juliusv:


juliusv:
@beorn7

bridge.go
```diff
@@ -1,8 +1,15 @@
-// Copyright (c) 2013, Prometheus Team
-// All rights reserved.
+// Copyright 2013 Prometheus Team
```
beorn7: This should be "The Prometheus Authors".

bridge_test.go
```diff
@@ -1,8 +1,15 @@
-// Copyright (c) 2013, Prometheus Team
-// All rights reserved.
+// Copyright 2013 Prometheus Team
```
beorn7: And again, and everywhere...

bridge.go
juliusv: Ah. Right. Then this needs to be fixed in the PGW as well, where I stole this header snippet from :)

juliusv:
Fixed throughout.

beorn7:
:+1:

PR #9, opened 2015-02-22
Use Makefile.COMMON for building statsd_bridge.

juliusv:


juliusv:
@grobie

grobie:
We should add a prominent banner to the Makefile.COMMON files explaining that they should not be touched and that the original file is in prometheus/utils.

juliusv:
@grobie Like this? https://github.com/prometheus/utils/pull/4

It's also a bit annoying that one needs to manually update the app repos after any chances there... would be great to get some automated propagation going (but yeah, so many other things to do)...

juliusv:
Updated the Makefile.COMMON here too.

juliusv:
I can haz :+1:?

juliusv:
Merging for now...

grobie:
:+1: sry

PR #10, opened 2015-04-10
Docker Support

ekristen:
This also allows for `-` to be in the metric name.

You can also publish a docker image to the `prom` organization now with this pull request.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,8 @@
+FROM golang:1.3-onbuild
+
+EXPOSE 9102
+EXPOSE 9125
+
+VOLUME ["/mapping.conf"]
+
+ENTRYPOINT ["/go/bin/app"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM golang:1.3-onbuild
 
 EXPOSE 9102
-EXPOSE 9125
+EXPOSE 9125/udp
 
 VOLUME ["/mapping.conf"]
 
--- a/README.md
+++ b/README.md
@@ -82,3 +82,13 @@ follows:
      
     test.web-server.foo.bar (gauge)
      => test_web__server_foo_bar_gauge{}
+
+## Docker
+
+```
+docker run -d --name="statsd_bridge" \
+  -p 9102:9102 \
+  -p 9125:9125 \
+  -v /path/on/host/to/mapping.conf:/mapping.conf \
+  statsd_bridge -statsd.mapping-config="/mapping.conf"
+```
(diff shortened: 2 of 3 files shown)


juliusv:
CC @discordianfish for the docker changes - I'm still looking at the regex ones to make sure there's no unintended side effects.

juliusv:
@ekristen So the `identifierRE` should actually be different in the different places where it's used: `metricLineRE`, `labelLineRE`, and `metricNameRE`:

- `metricLineRE`: this is about matching **StatsD metrics**, so any characters valid in StatsD metrics should be valid here.
- `labelLineRE`: this is about matching **Prometheus label names**, which cannot contain dashes and must match `[a-zA-Z_][a-zA-Z0-9_]*`.
- `metricNameRE` is about matching valid **Prometheus metric names**, which cannot contain dashes and must match `[a-zA-Z_:][a-zA-Z0-9_:]*`.

See also http://prometheus.io/docs/concepts/data_model/#metric-names-and-labels.

So I think the right thing to do here is to not reuse a single `identifierRE` for all of these, but limit each to the correct regexes only.

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM golang:1.3-onbuild
+
+EXPOSE 9102
+EXPOSE 9125/udp
+
+VOLUME ["/mapping.conf"]
```
discordianfish: I would omit this since it would force people to use a volume where I would rather create a custom image based on this one. You can still use a bind-mount even without this volume definition.

ekristen: I'm fine with removing it however, it doesn't force anyone to use a volume, just advertises that one can be placed there. Sorta a convenience item. The expose command is very similar, just advertises those ports they don't have to be used.

discordianfish: Actually this isn't even working since Volume definition in Dockerfile creates a directory, so if you don't bind mount you can't even place a file there:

```
FROM ubuntu
VOLUME [ "/foo.txt" ]
```

Build and run:
```
$ docker run -ti test bash
$ docker run test sh -c 'echo foo > /foo.txt'
sh: 1: cannot create /foo.txt: Is a directory
```

juliusv:
@ekristen Hey, any interest in continuing this? The metric matching code should probably be fixed to be similar/the same to https://github.com/prometheus/graphite_exporter/blob/master/mapper.go#L26-L31

discordianfish:
Also consider using alpine linux like we do now in other projects like prometheus server. This results in much smaller images.

ekristen:
I agree with alpine Linux. I haven't had time to dive into this anymore just yet unfortunately.

sdurrheimer:
@juliusv I think you can close this.

PR #12, opened 2015-04-16
Correct example, need to specify variables like in bash.

brian-brazil:
See http://golang.org/pkg/regexp/#Regexp.Expand


juliusv:
Can you add a test case for this in https://github.com/prometheus/statsd_bridge/blob/master/mapper_test.go?

juliusv:
:+1:

PR #22, opened 2015-09-09
Allow configuration of UDP read buffer

fstehle:
in order to avoid UDP packet loss when receiving a lot of packets on StatsD bridge.

PTAL: @juliusv


main.go
```diff
@@ -112,6 +113,14 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
+
+	if *readBuffer != 0 {
+		err = conn.SetReadBuffer(*readBuffer)
+		if err != nil {
+			log.Fatal(err)
```
juliusv: Could you do:

```go
log.Fatal("Error setting UDP read buffer:", err)
```

juliusv:
Thanks! :+1: otherwise

main.go
fstehle: I changed this.

juliusv:
Cool, thanks!

PR #24, opened 2015-10-08
allow metrics with dashes when mapping

macb:
It wasn't entirely clear why this wasn't allowed. We have some systems that submit metrics with their hostname in the metric. Something in the form of `test-myhost01`. This allows the mapper to match those.

Actually doesn't fully meet the use case where the name may be `test-my-host01` or something. Still working.

### DIFF
--- a/mapper.go
+++ b/mapper.go
@@ -24,8 +24,10 @@ import (
 )
 
 var (
-	identifierRE = `[a-zA-Z_][a-zA-Z0-9_]+`
-	metricLineRE = regexp.MustCompile(`^(\*\.|` + identifierRE + `\.)+(\*|` + identifierRE + `)$`)
+	identifierRE   = `[a-zA-Z_][a-zA-Z0-9_]+`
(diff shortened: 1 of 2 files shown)


mapper.go
```diff
@@ -24,7 +24,7 @@ import (
 )
 
 var (
-	identifierRE = `[a-zA-Z_][a-zA-Z0-9_]+`
+	identifierRE = `[a-zA-Z_]+-?[a-zA-Z0-9_]+`
```
juliusv: Thanks!

`identifierRE` is used not only in the regex that matches statsd metric lines, but also for the other two ones `labelLineRE` and `metricNameRE`. Those are for matching Prometheus labels and Prometheus metrics, both of which may not contain dashes.

If you want to allow dashes in statsd metric names, you'll have to isolate that change to just the regex that matches those.

macb: Will do!

macb:
ok @juliusv, teased apart the statsd metric line so that it could be changed independent of the labels regex.

juliusv:
@macb Cool, thanks!

PR #25, opened 2015-10-10
add root endpoint with redirect

macb:
Adds a root endpoint similar to [haproxy exporter](https://github.com/prometheus/haproxy_exporter/blob/master/haproxy_exporter.go#L391-L398) and [node exporter](https://github.com/prometheus/node_exporter/blob/master/node_exporter.go#L179-L187)


juliusv:
:+1: Thanks! Sorry for the slow response on this one, been traveling for VelocityConf and all...

PR #26, opened 2015-10-10
rename bridge to exporter

### LINKED ISSUE #13: Rename to statsd_exporter
To be consistent with the rest of our exporters, this repository should be renamed to the statsd_exporter as it takes in data from a non-prometheus system and exports it in a prometheus format.

By contrast a "bridge" would take data in some prometheus format, and output it in a non-prometheus format.

macb:
This would rename all relevant instances of `bridge` to `exporter`. This includes the bin itself and the docker container, which would both need to be updated. The repo would also need to be renamed (github will forward to the new repo for some amount of time fwiw).

closes #13


grobie:
Looks good to me. @juliusv wdyt?

juliusv:
:+1: Thanks!

PR #27, opened 2015-11-20
Update README.md

lowstz:
I can't pull the image use `docker pull prom/statsd-exporter`.
The `statsd-exporter` on docker hub is naming `prom/statsd-bridge` :whale:


grobie:
We should rather make the image available under the new name.

PR #28, opened 2015-12-01
Added support for sampling factor on timing events

yoshiyaka:
Added support for sampling rate in timings. This will for instance accept the statsd line

```
object-server.REPLICATE.timing:0.500917434692|ms|@0.1
```

The above line is real and was sent by openstack swift in our environment.

When a timing event with a sample rate set is received, it will be cloned a number of times in order to simulate it being received without a sampling rate set. This is kind of a dirty hack, but I think that it should be fine in practice.

### DIFF
--- a/bridge_test.go
+++ b/bridge_test.go
@@ -199,6 +199,21 @@ func TestHandlePacket(t *testing.T) {
 				},
 			},
 		}, {
+			name: "timings with sampling factor",
+			in:   "foo.timing:0.5|ms|@0.1",
+			out: Events{
+				&TimerEvent{metricName: "foo.timing", value: 0.5, labels: map[string]string{}},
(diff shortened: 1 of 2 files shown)


grobie:
We apparently missed this pull request. @yoshiyaka Can you rebase against current master and adjust?

sdurrheimer:
@yoshiyaka Are you still maintaining this PR ?

yoshiyaka:
Yes, but there are quite a few merge conflicts that I don't have the time to solve right now. Should I close this and open a new PR once they've been fixed?

SuperQ:
@yoshiyaka Ping, do you want to fix this up and get it merged?

yoshiyaka:
Now it should be ready to go. Sorry for the delay :)

grobie:
Thanks! We should refactor and clean up the statsd exporter at some point, but that's outside of the scope of this PR.

PR #31, opened 2016-02-01
fix documentation for docker udp port publishing

mmethner:
@juliusv Docker will fail with the current example run command.
```
$ docker run -d -p 9102:9102 -p 9125/udp:9125/udp prom/statsd-exporter
docker: Invalid hostPort: 9125/udp.
```


juliusv:
:+1: Thanks!

Issue #11, opened 2015-04-10
Bridge Crashes When Counter Decreases


ekristen:
Basically I'm forwarding `nsq` stats to statsd, with statsd repeating to the bridge. When NSQ restarts its stats restart, or when a NSQ reader disconnects and reconnects the stats reset, therefore the counter would reset.

I was told in IRC that Prometheus itself understands concepts like this. 

Is there a way we can make the bridge not crash in this scenario?

```
2015/04/10 23:17:12 Starting StatsD -> Prometheus Bridge...
2015/04/10 23:17:12 Accepting StatsD Traffic on :9125
2015/04/10 23:17:12 Accepting Prometheus Requests on :9102
panic: counter cannot decrease in value

goroutine 16 [running]:
runtime.panic(0x6fef00, 0xc20870fb10)
        /usr/src/go/src/pkg/runtime/panic.c:279 +0xf5
github.com/prometheus/client_golang/prometheus.(*counter).Add(0xc208196200, 0xc0ef876000000000)
        /go/src/github.com/prometheus/client_golang/prometheus/counter.go:69 +0xf9
main.(*Bridge).Listen(0xc20800f500, 0xc2080a2000)
        /go/src/app/bridge.go:217 +0x617
main.main()
        /go/src/app/main.go:127 +0x5e3

goroutine 19 [finalizer wait, 8 minutes]:
runtime.park(0x41a430, 0x952a78, 0x950cc9)
        /usr/src/go/src/pkg/runtime/proc.c:1369 +0x89
runtime.parkunlock(0x952a78, 0x950cc9)
        /usr/src/go/src/pkg/runtime/proc.c:1385 +0x3b
runfinq()
        /usr/src/go/src/pkg/runtime/mgc0.c:2644 +0xcf
runtime.goexit()
        /usr/src/go/src/pkg/runtime/proc.c:1445

goroutine 26 [IO wait, 5 minutes]:
net.runtime_pollWait(0x7f1489fffb20, 0x72, 0x0)
        /usr/src/go/src/pkg/runtime/netpoll.goc:146 +0x66
net.(*pollDesc).Wait(0xc2080ba140, 0x72, 0x0, 0x0)
        /usr/src/go/src/pkg/net/fd_poll_runtime.go:84 +0x46
net.(*pollDesc).WaitRead(0xc2080ba140, 0x0, 0x0)
        /usr/src/go/src/pkg/net/fd_poll_runtime.go:89 +0x42
[...]

PR #33, opened 2016-03-30
Don't panic when negative number counter is submitted.

klobucar:
Instead now we will send a warning message to the logs
and continue on with our day.

Also a guard has been added when we detect a channel has been
closed and the `Exporter.Listen` function will return out of
its loop vs always running. When a channel is closed there will
be no more data to consume, so this seems correct. It also allows
for a test to be written for the panic.

@juliusv @SuperQ

### DIFF
--- a/exporter_test.go
+++ b/exporter_test.go
@@ -1,0 +1,52 @@
+// Copyright 2013 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"testing"
+	"time"
+)
+
+// TestNegativeCounter validates when we send a negative
+// number to a counter that we no longer panic the Exporter Listener.
+func TestNegativeCounter(t *testing.T) {
+	defer func() {
+		if e := recover(); e != nil {
+			err := e.(error)
+			if err.Error() == "counter cannot decrease in value" {
+				t.Fatalf("Counter was negative and causes a panic.")
+			} else {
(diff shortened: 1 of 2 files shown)


klobucar:
This would fix #11

grobie:
Looks great! :+1:

SuperQ:
A couple ideas:
* When logging, log the current counter value as well.
* Instead of just continuing, use `counter.Set(0.0)` to reset the counter.

klobucar:
I think we should look at the behavior of statsd here a little more before we set it to (0.0) if we get a negative number.

Looking at how this code works and what statsd does I think we may be confusing a few data types.

 - Counters on the flush interval will turn to 0 (reading the statsd code they may support negative numbers, however Promethues doesn't)
 - Gauges do not reset on flush and allow for negative values
 - Sets are a one-off value that is replaced (which we have no support for currently in the code that I see)


From looking at these facts I'd rather warn and document we don't support counters de-incrementing and a gauge is a better data type to use for those cases.

SuperQ:
@klobucar Yea, good idea.  Let's go forward with logging first.  Unfortunately the counter object doesn't support `Get()`, so we can't log it.

SuperQ:
:+1: @jrv, care to take a peak?

exporter_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+package main
+
+import (
+	"testing"
+	"time"
+)
+
+// TestNegativeCounter validates when we send a negative
+// number to a counter that we no longer panic the Exporter Listener
```
juliusv: Tiny nit: end doc comment with period.

exporter_test.go
```diff
@@ -0,0 +1,52 @@
...
+func TestNegativeCounter(t *testing.T) {
+	defer func() {
+		if e := recover(); e != nil {
+			err := e.(error)
+			if err.Error() == "counter cannot decrease in value" {
+				t.Fatalf("Counter was negative and causes a panic.")
+			} else {
+				t.Fatalf("UNKNOWN PANIC AND ERROR: %q", err.Error())
```
juliusv: Nit: I'd just go with normal capitalization - no need for extra shouting IMO :)

juliusv:
:+1: otherwise, thanks!

SuperQ:
@klobucar Would you mind fixing the last couple of Nits?  Then I'll merge an release this.

klobucar:
Ok all nits taken care of.

juliusv:
:+1: Thanks :)

PR #34, opened 2016-04-23
retry: Support datadog extensions to statsd datagrams

fungusakafungus:
This is a rebased and improved version of https://github.com/prometheus/statsd_exporter/pull/3

We added some validation to DogStatsD tags and some tests. We also took it for a test run(it worked nicely) and plan to use it in production when this is merged.

We also added a special-case handling for the php dogstatsd client, which prefixes every tag with a `#`: https://github.com/DataDog/php-datadogstatsd/issues/35

I think that DogStatsD tags is a generally useful extension to StatsD protocol.

cc @pjjw @Bonko

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,8 @@ The exporter will convert DogStatsD-style tags to prometheus labels. See
 documentation for the concept description and
 [Datagram Format](http://docs.datadoghq.com/guides/dogstatsd/#datagram-format)
 for specifics. It boils down to appending
-`|#tag:value,another_tag:another_value` to the normal StatsD format.
+`|#tag:value,another_tag:another_value` to the normal StatsD format.  Tags
+without values (`#some_tag`) are not supported.
 
 ## Building and Running
 
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ In that case, you don't need to run a StatsD server anymore.
 We recommend this only as an intermediate solution and recommend switching to
 [native Prometheus instrumentation](http://prometheus.io/docs/instrumenting/clientlibs/)
 in the long term.
+
 ### DogStatsD extensions
 
 The exporter will convert DogStatsD-style tags to prometheus labels. See
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ The exporter will convert DogStatsD-style tags to prometheus labels. See
 documentation for the concept description and
 [Datagram Format](http://docs.datadoghq.com/guides/dogstatsd/#datagram-format)
 for specifics. It boils down to appending
-`#tag:value,another_tag:another_value` to the normal StatsD format.
+`|#tag:value,another_tag:another_value` to the normal StatsD format.
 
 ## Building and Running
 
--- a/README.md
+++ b/README.md
@@ -25,6 +25,14 @@ In that case, you don't need to run a StatsD server anymore.
 We recommend this only as an intermediate solution and recommend switching to
 [native Prometheus instrumentation](http://prometheus.io/docs/instrumenting/clientlibs/)
 in the long term.
+### DogStatsD extensions
+
+The exporter will convert DogStatsD-style tags to prometheus labels. See
+[Tags](http://docs.datadoghq.com/guides/dogstatsd/#tags) in the DogStatsD
+documentation for the concept description and
+[Datagram Format](http://docs.datadoghq.com/guides/dogstatsd/#datagram-format)
+for specifics. It boils down to appending
+`#tag:value,another_tag:another_value` to the normal StatsD format.
 
 ## Building and Running
 
--- a/bridge_test.go
+++ b/bridge_test.go
@@ -58,12 +58,32 @@ func TestHandlePacket(t *testing.T) {
 			},
 		}, {
 			name: "datadog tag extension",
+			in:   "foo:100|c|#tag1:bar,tag2:baz,tag3,tag4",
+			out: Events{
+				&CounterEvent{
+					metricName: "foo",
+					value:      100,
+					labels:     map[string]string{"tag1": "bar", "tag2": "baz", "tag3": ".", "tag4": "."},
+				},
+			},
+		}, {
+			name: "datadog tag extension with # in all keys (as sent by datadog php client)",
 			in:   "foo:100|c|#tag1:bar,#tag2:baz,#tag3,#tag4",
 			out: Events{
 				&CounterEvent{
 					metricName: "foo",
 					value:      100,
 					labels:     map[string]string{"tag1": "bar", "tag2": "baz", "tag3": ".", "tag4": "."},
+				},
+			},
+		}, {
+			name: "datadog tag extension with tags unsupported by prometheus",
+			in:   "foo:100|c|#09digits:0,tag.with.dots,tag_with_empty_value:",
+			out: Events{
+				&CounterEvent{
+					metricName: "foo",
+					value:      100,
+					labels:     map[string]string{"_09digits": "0", "tag_with_dots": ".", "tag_with_empty_value": "."},
 				},
 			},
 		}, {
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -25,6 +25,14 @@ In that case, you don't need to run a StatsD server anymore.
 We recommend this only as an intermediate solution and recommend switching to
 [native Prometheus instrumentation](http://prometheus.io/docs/instrumenting/clientlibs/)
 in the long term.
+### DogStatsD extensions
```
grobie: Let's keep an empty line between sections.

exporter.go
```diff
@@ -301,56 +311,84 @@ func (l *StatsDListener) handlePacket(packet []byte, e chan<- Events) {
...
-					log.Printf("Invalid sampling factor %s on line %s", samplingStr, line)
-					networkStats.WithLabelValues("invalid_sample_factor").Inc()
-					continue
-				}
-				if samplingFactor == 0 {
-					// This should never happen, but avoid division by zero if it does.
-					log.Printf("Invalid zero sampling factor %s on line %s, setting to 1", samplingStr, line)
-					samplingFactor = 1
+			if len(components) >= 3 {
+				for _, component := range components[2:] {
```
grobie: Let's extract that into it's own function. The indentation level is already pretty high. Something like

```go
if len(components) >= 3 {
  values := parseDataDogLabels(components[2:])
}
```

exporter.go
```diff
@@ -301,56 +311,84 @@ func (l *StatsDListener) handlePacket(packet []byte, e chan<- Events) {
...
 			components := strings.Split(sample, "|")
 			samplingFactor := 1.0
-			if len(components) < 2 || len(components) > 3 {
+			if len(components) < 2 || len(components) > 4 {
 				networkStats.WithLabelValues("malformed_component").Inc()
 				log.Println("Bad component on line:", line)
 				continue
 			}
 			valueStr, statType := components[0], components[1]
+			labels := map[string]string{}
```
grobie: Let's move this down above the len(components) check to keep related code together.

exporter.go
```diff
@@ -301,56 +311,84 @@ func (l *StatsDListener) handlePacket(packet []byte, e chan<- Events) {
...
+						}
+						value /= samplingFactor
+					case '#':
+						networkStats.WithLabelValues("dogstatsd_tags").Inc()
+						tags := strings.Split(component, ",")
+						for _, t := range tags {
+							t = strings.TrimPrefix(t, "#")
+							kv := strings.Split(t, ":")
+							tag_key := kv[0]
+							tag_key = escapeMetricName(tag_key)
```
grobie: Why the extra assignment and not just `tag_key = escapeMetricName(kv[0])`? It's also not usual to use long variable names for local variables in go.  Furthermore, golang variables use camelCase and not underscore. I'd just use `key`/`value`. The tag context is clear already.

exporter.go
grobie: Now I see, it's not only about Data dog labels. It's still be great to break up the code a bit here, 3 nested for loops + case and if statements is too much.

bridge_test.go
```diff
@@ -51,6 +53,47 @@ func TestHandlePacket(t *testing.T) {
...
+				},
+			},
+		}, {
+			name: "datadog tag extension with # in all keys (as sent by datadog php client)",
+			in:   "foo:100|c|#tag1:bar,#tag2:baz,#tag3,#tag4",
+			out: Events{
+				&CounterEvent{
+					metricName: "foo",
+					value:      100,
+					labels:     map[string]string{"tag1": "bar", "tag2": "baz", "tag3": ".", "tag4": "."},
```
grobie: I'm thinking whether `.` makes it clear that this is a key only label. What about something like `<n/a>`?

Also, is it common that implementation use many different varying tags here? This would be problematic, as each unique label set will result in a new time series. It's also not allowed to use metrics with varying label sets in general.

pjjw:
awesome! always meant to take another crack at this.

bridge_test.go
fungusakafungus: `<n/a>` or dot: Wouldn't `<n/a>` make querying such metrics a bit awkward? Like this `http_requests_total{live="<n/a>"}`

Besides, I'm yet to find a single useful example of a valueless tag in datadog docs, maybe we shouldn't implement this at all? Ok, now it sounds like bikeshedding. I'll change it.

fungusakafungus: Your second point: the implementer decides on the tags. Same problem as having many variable parts in the metric name, like `stats_counts.production.web67.request_time.signup.ab_test_13.cancelled`. Yes, you can shoot yourself in the foot with it. No, you shouldn't.

fungusakafungus:
Please take a look again, review changes implemented.

bridge_test.go
```diff
@@ -51,6 +53,47 @@ func TestHandlePacket(t *testing.T) {
...
+				},
+			},
+		}, {
+			name: "datadog tag extension",
+			in:   "foo:100|c|#tag1:bar,tag2:baz,tag3,tag4",
+			out: Events{
+				&CounterEvent{
+					metricName: "foo",
+					value:      100,
+					labels:     map[string]string{"tag1": "bar", "tag2": "baz", "tag3": "<n/a>", "tag4": "<n/a>"},
```
grobie: I'm still unsure about the tag support. Are you using that in your setup? I'd assume these are rarely static names with only one possible value. Querying metrics with varying label sets (different label keys) will result in all kind of problems.

I'd suggest to either forbid the usage of tags or use them as value instead. For example:

```
# input
foo:100|c|#a:b,foo,bar
foo:300|c|#a:d,qux,bar

# metrics
foo{a="b", tag2="foo", tag3="bar"} 100
foo{a="b", tag2="qux", tag3="bar"} 300
```

This still leaves the problem that the caller has to ensure the order is stable.

pjjw: unfortunately that's not how datadog does it, and you kinda have to preserve this somehow if you want to allow people to migrate from datadog to prom

grobie: I'm not sure I understand what you mean. The point is that the current suggestion is invalid with the Prometheus model in case these positional tags have different values. The only option I see is to use a fake label name like `tag{number}`.

fungusakafungus:
On valueless tags: we(my company) are not using them and are not going to. Datadog documentation mentions them, but fails to provide a single useful example or screenshot where they are used. I'd say we shouldn't care about them.

I myself have two different exciting proposals on how to implement them, but I'd prefer to keep those to myself or suggest them to anyone who happens to miss this feature. Let's ignore the valueless tags until someone using them tells us what to do with them. 

Tags in the form of `|#name:value` are perfectly useful, and we intend to them.

@pjjw: are you using valueless tags? Any specifics?

Migrating from DataDog to prometheus is not the purpose of this pull request. Supporting one particular useful StatsD extension is.

The use is to move specifying labels away from `-statsd.mapping-config` into clients, while still using a lightweight UDP protocol.

pjjw:
yeah fair point, not using them. silent drop would probably be fine.

pjjw:
more the reason I chimed in was to be against tag1= tag2= proposal, dropping is better

grobie:
This makes sense to me.  The idea of generating label names was just the alternative suggestion in case tag support is a requirement. 

* please remove the support for valueless tags.
* optionally: print a warning when such a metric is sent (might be too spammy)
* add a note to the readme that callers have to ensure to always send the same label sets for a given metric

fungusakafungus:
> please remove the support for valueless tags.

Done that.

> optionally: print a warning when such a metric is sent (might be too spammy)

That too, and added a metric for malformed DogStatsD tags.

> add a note to the readme that callers have to ensure to always send the same label sets for a given metric

I'm not comfortable writing this, as I don't quite understand the effects... You might want to provide me with a proper sentence/paragraph or commit it yourself later.

grobie:
Awesome! Looks great :+1: @juliusv Any last comments?

Please squash the commits together.

juliusv:
:+1: thanks! Could you please squash the commits? Then we'll merge.

fungusakafungus:
pull-requested the squashed version.

PR #37, opened 2016-05-02
Make metric suffixes optional, expose unmapped events

discordianfish:
I change the README to show all examples without the auto suffixes but mention that, for now, they still get added. I did that because I think in one of the next releases we should change the defaults: Even though it might break things, I believe all our exporters should be serve as idiomatic example which here isn't the case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,28 +57,33 @@ Metrics that don't match any mapping in the configuration file are translated
 into Prometheus metrics without any labels and with certain characters escaped
 (`_` -> `__`; `-` -> `__`; `.` -> `_`).
 
-In general, the different metric types are translated as follows, with certain
-suffixes appended to the Prometheus metric names:
+In general, the different metric types are translated as follows:
 
-    StatsD gauge   -> Prometheus gauge (suffix `_gauge`)
+    StatsD gauge   -> Prometheus gauge
 
-    StatsD counter -> Prometheus counter (suffix `_counter`)
+    StatsD counter -> Prometheus counter
 
-    StatsD timer   -> Prometheus summary (suffix `_timer`)        <-- indicates timer quantiles
-                   -> Prometheus counter (suffix `_timer_total`)  <-- indicates total time spent
-                   -> Prometheus counter (suffix `_timer_count`)  <-- indicates total number of timer events
+    StatsD timer   -> Prometheus summary                    <-- indicates timer quantiles
+                   -> Prometheus counter (suffix `_total`)  <-- indicates total time spent
+                   -> Prometheus counter (suffix `_count`)  <-- indicates total number of timer events
 
-An example mapping configuration:
+If `-statsd.add-suffix` is set, the exporter appends the metric type (`_gauge`,
+`_counter`, `_timer`) to the resulting metrics. This is enabled by default for
+backward compatibility but discouraged to use. Instead, it is better to
+explicitly define the full metric name in your mapping and run the exporter
+with `-statsd.add-suffix=false`.
+
+An example mapping configuration with `-statsd.add-suffix=false`:
 
     test.dispatcher.*.*.*
-    name="dispatcher_events"
+    name="dispatcher_events_total"
     processor="$1"
     action="$2"
     outcome="$3"
     job="test_dispatcher"
 
     *.signup.*.*
-    name="signup_events"
+    name="signup_events_total"
     provider="$2"
     outcome="$3"
     job="${1}_server"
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -57,28 +57,33 @@ Metrics that don't match any mapping in the configuration file are translated
...
 
-    StatsD timer   -> Prometheus summary (suffix `_timer`)        <-- indicates timer quantiles
-                   -> Prometheus counter (suffix `_timer_total`)  <-- indicates total time spent
-                   -> Prometheus counter (suffix `_timer_count`)  <-- indicates total number of timer events
+    StatsD timer   -> Prometheus summary                    <-- indicates timer quantiles
+                   -> Prometheus counter (suffix `_total`)  <-- indicates total time spent
+                   -> Prometheus counter (suffix `_count`)  <-- indicates total number of timer events
 
-An example mapping configuration:
+If `-statsd.add-suffix` is set, the exporter appends the metric type (gauge,
```
juliusv: Be more explicit/precise about what they can expect: "(`_gauge`, `_counter`, `_timer`)".

README.md
```diff
@@ -57,28 +57,33 @@ Metrics that don't match any mapping in the configuration file are translated
...
+counter, timer) to the resulting metrics. This is enabled by default for
+backward compatibility but discouraged to use. Instead better explicitly define
```
juliusv: "Instead, it is better to explicitly define..."

README.md
```diff
@@ -87,13 +92,13 @@ This would transform these example StatsD metrics into Prometheus metrics as
 follows:
 
     test.dispatcher.FooProcessor.send.success (counter)
```
juliusv: I think you can get rid of the "(counter)" and "(gauge)" comments now that they don't actually have an effect in the example anymore.

main.go
```diff
@@ -30,6 +30,7 @@ var (
 	statsdListenAddress = flag.String("statsd.listen-address", ":9125", "The UDP address on which to receive statsd metric lines.")
 	mappingConfig       = flag.String("statsd.mapping-config", "", "Metric mapping configuration file name.")
 	readBuffer          = flag.Int("statsd.read-buffer", 0, "Size (in bytes) of the operating system's transmit read buffer associated with the UDP connection. Please make sure the kernel parameters net.core.rmem_max is set to a value greater than the value specified.")
+	addSuffix           = flag.Bool("statsd.add-suffix", true, "Add the metric type (counter/gauge/timer) as suffix to the generated Prometheus metric (NOT recommended).")
```
juliusv: Maybe say "(NOT recommended, but set by default for backward compatibility)"?

main.go
```diff
@@ -108,6 +109,9 @@ func watchConfig(fileName string, mapper *metricMapper) {
 func main() {
 	flag.Parse()
 
+	if *addSuffix {
+		log.Println("Warning: Using -statsd.add-suffix is discouraged. Better explicitly name metrics in config according to best pratices.")
```
juliusv: Instead of "Better..." I'd go with "We recommend explicitly naming metrics appropriately in the mapping configuration."

exporter.go
```diff
@@ -189,6 +189,14 @@ func escapeMetricName(metricName string) string {
 	return metricName
 }
 
+func suffix(metricName, suffix string) string {
+	str := metricName
+	if *addSuffix {
```
juliusv: Let's not make this code base even uglier by directly reading flag values from here.

Make it an `Exporter` method instead and pass the desired value into `NewExporter`?

telemetry.go
```diff
@@ -25,6 +25,10 @@ var (
 		},
 		[]string{"type"},
 	)
+	eventsMissCount = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "statsd_exporter_events_misses_total",
```
juliusv: I'd go for `statsd_exporter_unmapped_events_total`, easier to understand what it's about.

juliusv: (and then the var name should be renamed accordingly)

telemetry.go
```diff
@@ -40,8 +44,8 @@ var (
 		[]string{"outcome"},
 	)
 	mappingsCount = prometheus.NewGauge(prometheus.GaugeOpts{
-		Name: "statsd_exporter_loaded_mappings_count",
-		Help: "The number of configured metric mappings.",
+		Name: "statsd_exporter_loaded_mappings",
```
juliusv: While this is a better metric name and I support it, can we keep this breaking unrelated change out of this PR? Otherwise we will probably not catch it for the Changelog entry and all.

juliusv:
:+1: otherwise

exporter.go
```diff
@@ -209,13 +218,14 @@ func (b *Exporter) Listen(e <-chan Events) {
...
 			} else {
+				eventsUnmapped.Inc()
 				metricName = escapeMetricName(event.MetricName())
 			}
 
 			switch event.(type) {
 			case *CounterEvent:
 				counter := b.Counters.Get(
-					metricName+"_counter",
+					b.suffix(metricName, "counter"),
```
SuperQ: While we're fixing this, can we change this to `total` to match Prometheus metric name guidelines?

juliusv: @SuperQ The whole idea of introducing the `-statsd.add-suffix` flag (instead of just always omitting automatic suffixes) was to not have a breaking change. Changing the suffix to `_total` here would be breaking though.

juliusv:
:+1:

PR #41, opened 2016-05-03
Drop _count suffix for loaded_mappings metric 

discordianfish:
This makes the naming consistent with our best pratices.


juliusv:
:+1:

PR #42, opened 2016-05-03
New build process

### LINKED ISSUE #39: Add the new release process

discordianfish:
This closes #39


juliusv:
@sdurrheimer

.promu.yml
```diff
@@ -0,0 +1,25 @@
+go: 1.5.4
```
sdurrheimer: Given that Prometheus and Alertmanager are now `1.6.2`, I think we can change here too.

.promu.yml
```diff
@@ -0,0 +1,25 @@
...
+crossbuild:
+    platforms:
+        - linux/amd64
+        - linux/386
+        - darwin/amd64
+        - darwin/386
+        - windows/amd64
+        - windows/386
+        - netbsd/amd64
+        - netbsd/386
```
sdurrheimer: Are these the full working platforms for `statsd_exporter` ?

.travis.yml
```diff
@@ -0,0 +1,9 @@
+sudo: false
+
+language: go
+go:
+- 1.5
```
sdurrheimer: `1.5.4` and `1.6.2` would be better.

Dockerfile
```diff
@@ -1,4 +1,7 @@
-FROM        sdurrheimer/alpine-golang-make-onbuild
+FROM        quay.io/prometheus/busybox:glibc
```
sdurrheimer: Is the `linux/amd64` binary for `statsd_exporter` not statically buildable ?

.promu.yml
discordianfish: Fixed!

README.md
```diff
@@ -1,5 +1,9 @@
-statsd_exporter
-=============
+# statsd exporter [![Build Status](https://travis-ci.org/prometheus/statsd_exporter.svg)][travis]
+
+[![CircleCI](https://circleci.com/gh/prometheus/statsd_exporter/tree/master.svg?style=shield)][circleci]
+[![Image Size & Layers](https://imagelayers.io/badge/prom/statsd-exporter:master.svg)][imagelayers]
+[![Docker Stars](https://img.shields.io/docker/stars/prom/statsd-exporter.svg)][hub]
+[![Docker Pulls](https://img.shields.io/docker/pulls/prom/statsd-exporter.svg)][hub]
```
sdurrheimer: The shields discussion is [still going](https://github.com/prometheus/prometheus/pull/1598), but I think we should replace the `imagelayer` shield by a Quay link and remove the `Docker Stars`.

```
[![Docker Repository on Quay](https://quay.io/repository/prometheus/statsd-exporter/status)][quay]

[quay]: https://quay.io/repository/prometheus/statsd-exporter
```

.travis.yml
discordianfish: Both? Or just 1.6.2? I simply used the node-exporter as example. I'll change it for now to 1.6.2.

discordianfish: Just tested it and they can all be built. Do we want to support more than those?

Dockerfile
discordianfish: Changed it to busybox:latest.

.travis.yml
sdurrheimer: People might still build with `go 1.5`, so I would build with `1.5.4` and `1.6.2`.
I don't really have a opinion about `tip`, is a `devel` version really useful ?

.promu.yml
discordianfish: Took that from node-exporter. I've tried and they all can be built. Do we want to support more/less platforms?

sdurrheimer: Given that there a lot of release, we can do the full list if they all works:

```
    platforms:
        - linux/amd64
        - linux/386
        - darwin/amd64
        - darwin/386
        - windows/amd64
        - windows/386
        - freebsd/amd64
        - freebsd/386
        - openbsd/amd64
        - openbsd/386
        - netbsd/amd64
        - netbsd/386
        - dragonfly/amd64
        - linux/arm
        - linux/arm64
        - freebsd/arm
        - openbsd/arm
        - netbsd/arm
        - linux/ppc64
        - linux/ppc64le
        - linux/mips64
        - linux/mips64le
```

grobie: If we support all platforms, let's not specify it here in promu and use the defualt. then we can add new versions without the need to touch x repositories.

sdurrheimer: Good point 👍

sdurrheimer:
👍

PR #50, opened 2016-07-20
Skip metrics with invalid utf8

fungusakafungus:
This would prevent an incident I have been seeing:

The exporter will happily store metrics with invalid utf8 in labels. Prometheus will then try to collect them and will choke on invalid utf8, [here](https://github.com/prometheus/prometheus/blob/e83f05fe93c32c9e808b2a6a907337b442a540dc/vendor/github.com/prometheus/common/model/labels.go#L157) and [here](https://github.com/prometheus/prometheus/blob/dc4d5fef2a75b9b4e5124fe41ec291eda52f667e/vendor/github.com/prometheus/common/expfmt/decode.go#L104) I think. The statsd-exporter target will be  marked down because of this error.

I wonder if the `utf8.ValidString` check for every input line would make StatsD ingestion considerably slower/increase CPU usage?

I've tried to only check DogStatsD tag values first, but it wasn't enough and it was too much code. Which explains the many new tests.

TODO
 - [x] remove unintended changes

### DIFF
--- a/bridge_test.go
+++ b/bridge_test.go
@@ -137,6 +137,22 @@ func TestHandlePacket(t *testing.T) {
 				},
 			},
 		}, {
+			name: "datadog tag extension with invalid utf8 tag values",
+			in:   "foo:100|c|@0.1|#tag:\xc3\x28invalid",
+		}, {
+			name: "datadog tag extension with both valid and invalid utf8 tag values",
+			in:   "foo:100|c|@0.1|#tag1:valid,tag2:\xc3\x28invalid",
+		}, {
+			name: "multiple metrics with invalid datadog utf8 tag values",
+			in:   "foo:200|c|#tag:value\nfoo:300|c|#tag:\xc3\x28invalid",
+			out: Events{
+				&CounterEvent{
+					metricName: "foo",
+					value:      200,
+					labels:     map[string]string{"tag": "value"},
+				},
+			},
+		}, {
 			name: "combined multiline metrics",
 			in:   "foo:200|ms:300|ms:5|c|@0.1:6|g\nbar:1|c:5|ms",
 			out: Events{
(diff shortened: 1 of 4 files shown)


fungusakafungus:
@grobie would you please take a look?

grobie:
For robustness, I think it makes sense to handle this on the statsd_exporter side, as a rogue client can otherwise prevent the ingestion of all statsd metrics. I have the same question you already asked though, how big is the performance impact here?

Would you mind writing a very simple benchmark (can totally be added to the test suite) and paste the number with+without this change? I'll check the usual throughput of statsd messages in our exporters, but I think it will be rather small. If you have some numbers from your systems, please share so that we can reason about the impact of this change.

Thanks again for your contribution!

fungusakafungus:
Doesn't affect performance (or I couldn't measure it properly):

```
(drop_invalid_utf8)$ go test -bench=. -benchtime=10s 2>/dev/null
PASS
BenchmarkExporter1-4      500000             48768 ns/op
BenchmarkExporter5-4       10000           1166882 ns/op
BenchmarkExporter50-4        100         104472505 ns/op
ok      github.com/prometheus/statsd_exporter   47.374s
(drop_invalid_utf8)$ git checkout master
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
(master)$ go test -bench=. -benchtime=10s 2>/dev/null
PASS
BenchmarkExporter1-4      300000             52192 ns/op
BenchmarkExporter5-4       10000           1226481 ns/op
BenchmarkExporter50-4        100         106328200 ns/op
ok      github.com/prometheus/statsd_exporter   39.460s
```

(tested with exporter_benchmark_test.go not git-added yet)

fungusakafungus:
In case it was not clear, the benchmark is committed, see 237499e.

grobie:
Awesome, thanks so much! I'll keep this PR in mind for exception contributions, very well tested!

Issue #53, opened 2016-10-11
Issues with exposed metrics


tcolgate:
statsd_exporter_packets_total does not count packets. It counts events, and the dimensions are not used consistantly. dogstatsd metrics get double counted as both legal and dogstatsd.

networkStats.WithLabelValues("legal").Inc() should be moved out of the per-line loop.
I can PR this, but the other counters need some thought too, the various forms of invalid should probably be on events (lines), rather than packets.

matthiasr:
This was fixed in #106.

PR #54, opened 2016-10-11
Fix packet counting

tcolgate:


tcolgate:
Fixes packet counts for #53 
The other dimensions are still dubious.
This also includes a fmt fix.

grobie:
Hm, yeah. We increase the counter multiple times per packet in case of malformed/unexpected payload. I guess all the errors should be extracted into other metrics before this can be considered "fixed".

SuperQ:
Ping, is still useful?  It needs a rebase.

matthiasr:
You are right, the other dimensions are very dubious. Even so, calling it a "packet count" does not really do it justice. I went through and reworked everything previously covered by this metric in #106.

Since this PR has been inactive for a long time, and #106 is a superset of what it is trying to do, I'll go ahead and close this one.

Issue #55, opened 2016-10-21
Not having a terminating new-line in statsd-mappings.conf makes last entry invalid


carlpett:
If there is no new-line after mappings, the last mapping is invalid. For example, this configuration file:

```
logstash.*.*.*.events
name="logstash_events_total"
source="$1"
instance="$2"
stage="$3"<no newline here>
```

... will not work. Adding the newline, it starts to work.

PR #61, opened 2017-02-22
Improve mapping configuration parser

vtorhonen:
Issue #55 shows that the mapping config is not parsed in full if a terminating newline is missing. The last metric mapping is not parsed at all and no errors are shown. This PR changes the parsing method by treating the last line of the file as if it would be a newline between two mappings.

Added two new testcases to `mapper_test.go`:

- a single mapping configuration without a terminating newline
- two mapping configurations without a terminating newline

Tested by running `go test` and it gives a big thumbs up. In addition I tested this locally by running `prom/statsd-exporter` (Dockerhub) and the fixed version side-by-side with this mapping:

```
*.test.*
name="test"
foo="$1"
bar="$2 <no newline>
```

Results for the official image:

```
$ curl -s 127.0.0.1:9102/metrics | grep ^statsd_exporter_build_info
statsd_exporter_build_info{branch="master",goversion="go1.6.2",revision="e00d9e8",version="0.3.0"} 1
$ (curl -s 127.0.0.1:9102/metrics | grep ^test) || echo ":("
:(
$ echo "foo.test.bar:1|g" >/dev/udp/127.0.0.1/9125
$ (curl -s 127.0.0.1:9102/metrics | grep ^test) || echo ":("
:(
```

Results after applying this PR:

```
$ curl -s 127.0.0.1:9102/metrics | grep ^statsd_exporter_build_info
statsd_exporter_build_info{branch="",goversion="go1.6.4",revision="",version=""} 1
$ (curl -s 127.0.0.1:9102/metrics | grep ^test) || echo ":("
:(
$ echo "foo.test.bar:1|g" >/dev/udp/127.0.0.1/9125
$ (curl -s 127.0.0.1:9102/metrics | grep ^test) || echo ":("
test{bar="bar",foo="foo"} 1
```

Pinging @juliusv as instructed in `CONTRIBUTIONS.md`.

### DIFF
--- a/mapper.go
+++ b/mapper.go
@@ -77,19 +77,10 @@ func (m *metricMapper) initFromString(fileContents string) error {
 			state = METRIC_DEFINITION
 
 		case METRIC_DEFINITION:
-			emptyLine, lastLine := false, false
+			if (i == numLines-1) && (line != "") {
+				return fmt.Errorf("Line %d: missing terminating newline", i)
+			}
 			if line == "" {
-				emptyLine = true
-			}
-			if i == numLines-1 {
-				lastLine = true
-			}
-			if (emptyLine) || (lastLine) {
-				if !emptyLine {
-					if err := m.updateMapping(line, i, &currentMapping); err != nil {
-						return err
-					}
-				}
 				if len(currentMapping.labels) == 0 {
 					return fmt.Errorf("Line %d: metric mapping didn't set any labels", i)
 				}
--- a/mapper.go
+++ b/mapper.go
@@ -51,6 +51,7 @@ const (
 
 func (m *metricMapper) initFromString(fileContents string) error {
 	lines := strings.Split(fileContents, "\n")
+	numLines := len(lines)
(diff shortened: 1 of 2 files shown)


grobie:
Instead of adding support to handle files without trailing newlines, I'd rather keep the current format but return an error if it's missing. It's the philosophy of Prometheus to only provide one way to do things.

vtorhonen:
Changed the behaviour as you suggested. When a terminating newline is missing an error such as this is shown:

```
time="2017-02-24T09:29:41Z" level=fatal msg="Error loading config:Line 3: missing terminating newline" source="main.go:158"
```

Modified tests and made sure they are working as expected.

While I do disagree with this design it's still better than before (being nothing). Usability-wise the original solution would've been better.

grobie:
Thanks a lot for your contribution!

Please follow the git commit message best practices, as described here for example: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

Issue #64, opened 2017-03-09
Gauges do not work according to Promethius/Statsd specs


GentlemanJ:
Gauges should provide functionality to increase/decrease the metric by the specified value, not just always set the metric to the value. 

For instance;

myGauge:1|g   should set this gauge to 1
myGauge:+1|g should increment the gauge by 1
myGauge:-1|g should decrement the gauge by 1

drawks:
This issue should be marked as resolved with the merge in #65 and the release cut subsequently in #69

PR #65, opened 2017-03-09
Added increment/decrement support to Gauges

GentlemanJ:
@juliusv 

Addresses Issue #64

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -384,9 +384,7 @@ func (l *StatsDListener) handlePacket(packet []byte, e chan<- Events) {
 			valueStr, statType := components[0], components[1]
 
 			var relative = false
-			if strings.Index(valueStr, "+") == 0 {
-				relative = true
-			} else if strings.Index(valueStr, "-") == 0 {
+			if strings.Index(valueStr, "+") == 0 || strings.Index(valueStr, "-") == 0 {
 				relative = true
 			}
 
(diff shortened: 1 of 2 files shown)


exporter.go
```diff
@@ -287,24 +303,27 @@ type StatsDListener struct {
...
 }
 
-func buildEvent(statType, metric string, value float64, labels map[string]string) (Event, error) {
+func buildEvent(statType, metric string, value float64, valueStr string, labels map[string]string) (Event, error) {
 	switch statType {
 	case "c":
 		return &CounterEvent{
 			metricName: metric,
 			value:      float64(value),
+			valueStr:   valueStr,
```
juliusv: Do we need to store the `valueStr` in all the metric types, or would it be sufficient to just add a `relative` boolean flag only to the `GaugeEvent` type and then just do this:

```go
if event.relative {
    gauge.Add(event.Value())
} else {
  gauge.Set(event.Value())
}
```

GentlemanJ: Yup, that sounds like a much better implementation. I'll do that

juliusv:
@SuperQ It's common practice in Go to omit a field in a struct initialization if you are just going to set it to its zero value (`false`), so I think it's fine as it is.

SuperQ:
Ahh, ok, makes sense.

Issue #19, opened 2015-07-23
Usage of histogram for timer metrics


marcusmartins:
Currently, the statsd_bridge uses summary as the mapping for statsd timers and I am wondering what you think about allowing the bridge to decide between using summary or histogram for statsd timers?

Some backstory:
As our clients are already instrumented with statsd, we're considering using the statsd_bridge as an initial step to migrate our webservices metrics into prometheus and validate our setup.

Long term, we plan to use the python-client (and contribute to it) but in the very short term I am somewhat blocked on how to deal with pre-fork servers like gunicorn in a clean way. The statsd_bridge would provide us with an IPC mechanism to send metrics from multiple instance/processes and expose it to prometheus. With the histogram support, that would allow us to run statsd_bridge per instance and expose a per instance metrics endpoint (a very desirable feature).

If that's a direction that is acceptable, a few things will need to be designed/implemented:
- Are "summary vs histogram" a runtime configuration or something that could be set on a per mapping basis?
- If the choice can be made on per mapping basis, should that be extended to handle bucket sizes or use a default set of buckets?

To get more familiar with the code, I made a small test branch to validate the histogram implementation:
https://github.com/prometheus/statsd_bridge/compare/master...marcusmartins:histogram_support?expand=1

juliusv:
At an offsite right now, but :+1: in general to using/allowing histograms instead of summaries. For histograms it's quite important to be able to configure the right bucket sizes though, as otherwise they can become quite useless. So that probably needs to be a per-mapping configuration. Not sure about how to integrate that with the configuration mapping language, but open for suggestions!

marcusmartins:
Looking at the current implementation of 'name', labels could be be used to indicated bucket arrays and the type of metric to be used. 

Something like:
```
request_duration_microseconds.*.*
name="request_duration_microseconds"
type="histogram"
buckets="5, 10, 25, 50, 100, 250, 500, 1000, 1500, 2500"
view="$1"
method="$2"
job="hub-web"
```

As buckets and type were not reserved words before, it'd not be a backwards compatible change. A work around would be to use __buckets and __type.

Another option would be to steal from the plain text format and do something like:
```
# TYPE request_duration_microseconds histogram 
# BUCKETS 5, 10, 25, 50, 100, 250, 500, 1000, 1500, 2500
request_duration_microseconds.*.*
view="$1"
method="$2"
job="hub-web"
```
[...]

PR #66, opened 2017-03-13
Allow histograms for timer metrics

bakins:
Based on comments in https://github.com/prometheus/statsd_exporter/issues/19 this allows the use of a YAML config file that allows setting histograms for timer metrics per mapping or as the default timer type. Previous behavior is the same when using the "old" config file format.

This is fairly rough and could probably be cleaner.  I did the minimum possible to enable histograms as an experiment.  Consider this enough to start a conversation.


bakins:
Anything I can do to help on this? We've been running my fork for a while and it seems to be working fine.

gerricom:
@bakins: Could you provide a sample configuration file? I do not understand how to map the labels with your yaml-solution... 😕

bakins:
@gerricom sorry for the delay.  

This PR was to start the discussion. I can add documentation if people think this is useful.

Example:

"normal" config:

```
test.dispatcher.*.*.*
name="dispatcher_events_total"
processor="$1"
action="$2"
outcome="$3"
job="test_dispatcher"
```

YAML version of this:
```yaml
mappings:
- match: test.dispatcher.*.*.*
  labels:
    name: "dispatcher_events_total"
    processor: "$1"
    action: "$2"
    outcome: "$3"
    job: "test_dispatcher"
```

To enable histograms for a single mapping:
```yaml
mappings:
- match: test.dispatcher.*.*.*
  timer_type: histogram
  labels:
    name: "dispatcher_events_total"
...
```

To enable histograms by default:
```yaml
defaults: 
  timer_type: histogram
```

gerricom:
@bakins No problem! Thanks for the example. I must have simply missed the match parameter in the YAML configuration.

So after playing around with it: This looks quite useful and I'd like to see this merged. Changing over to the YAML configuration style is very easily done and straightforward. Histograms seem to work as expected.

juliusv:
@bakins Sorry for ignoring this for so long. I think in general changing this to YML is a great idea, to get rid of yet another custom language. Taking a closer look now.

mapper.go
```diff
@@ -102,40 +113,90 @@ func (m *metricMapper) initFromString(fileContents string) error {
...
 	defer m.mutex.Unlock()
-	m.mappings = parsedMappings
+	m.Mappings = parsedMappings
 
 	mappingsCount.Set(float64(len(parsedMappings)))
 
 	return nil
 }
 
+func (m *metricMapper) initFromYamlString(fileContents string) error {
```
juliusv: style: Yaml -> YAML (https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

mapper_test.go
```diff
@@ -148,7 +148,7 @@ func TestMetricMapper(t *testing.T) {
 				test.bar
 				name="name_bar"
 				label="foo"
-				
+
 				test.foo
 				name="name_foo"
 				label="bar"`,
```
juliusv: Would be great to have some YAML-based tests here of course.

exporter.go
```diff
@@ -263,11 +299,28 @@ func (b *Exporter) Listen(e <-chan Events) {
...
-				summary.Observe(event.Value())
+				timerType := ""
+				if mapping != nil {
+					timerType = mapping.TimerType
+				}
+				if timerType == "" {
+					timerType = b.mapper.Defaults.TimerType
+				}
+
+				if timerType == "histogram" {
```
juliusv: As far as I can see, the timer type string isn't validated at config load time, but probably should be?

And then I'd change this to a `switch timerType {...}` here with a default case that panics. Could be good to extract the strings into named constants too.

exporter.go
```diff
@@ -263,11 +299,28 @@ func (b *Exporter) Listen(e <-chan Events) {
...
+					histogram := b.Histograms.Get(
+						b.suffix(metricName, "timer"),
```
juliusv: Hmm, I think we should only suffix metrics that don't have an explicitly mapped metric name (if there is a mapping, use exactly the metric name provided in the mapping). But I see we also do that for gauges and counters already. So that should probably be fixed in a separate change.

bakins:
This now validates the timer type when it reads the configuration. Also addresses a couple of the style issues. Let me know what you think. I'll add docs to this PR after we think the code looks good.

mapper.go
```diff
@@ -107,40 +118,92 @@ func (m *metricMapper) initFromString(fileContents string) error {
...
 
 	mappingsCount.Set(float64(len(parsedMappings)))
 
 	return nil
 }
 
+func (m *metricMapper) initFromYAMLString(fileContents string) error {
+	var n metricMapper
+
+	fmt.Println(fileContents)
```
juliusv: Remove debug Println here and Printf below.

exporter.go
```diff
@@ -273,17 +309,44 @@ func (b *Exporter) Listen(e <-chan Events) {
...
+				if t == timerTypeDefault {
+					t = b.mapper.Defaults.TimerType
+				}
 
-					eventStats.WithLabelValues("timer").Inc()
-				} else {
-					log.Errorf(regErrF, metricName, err)
-					conflictingEventStats.WithLabelValues("timer").Inc()
+				switch t {
+				case "histogram":
```
juliusv: `"histogram"` -> `timerTypeHistogram`?

grobie:
Could you add some tests for the new feature?

bakins:
Added basic config tests.

drawks:
This looks like a very nice change. One critique of mine would be to separate the conversion to yaml config and the addition of histograms as two separate PRs. I'm a big fan of both changes though personally.

mapper.go
```diff
@@ -107,40 +118,105 @@ func (m *metricMapper) initFromString(fileContents string) error {
...
+	}
+
+	for i := range n.Mappings {
+		currentMapping := &n.Mappings[i]
+
+		if !metricLineRE.MatchString(currentMapping.Match) {
+			return fmt.Errorf("invalid match: %s", currentMapping.Match)
+		}
+
+		// check that label is correct
```
juliusv: nit: start all comments upper-cased, end with period.

mapper.go
```diff
@@ -107,40 +118,105 @@ func (m *metricMapper) initFromString(fileContents string) error {
...
+		for k, v := range currentMapping.Labels {
+			label := fmt.Sprintf("%s=%q", k, v)
+			if len(labelLineRE.FindStringSubmatch(label)) != 3 {
```
juliusv: It is a bit complicated to first make a k=v string out of this and then match the entire string against `labelLineRE`. The label value can be anything, so it would be better to just check whether the label name (`k`) matches `identifierRE`.

mapper.go
```diff
@@ -107,40 +118,105 @@ func (m *metricMapper) initFromString(fileContents string) error {
...
+				return fmt.Errorf("invalid label: %s: %s", k, v)
+			}
+			if k == "name" && !metricNameRE.MatchString(v) {
+				return fmt.Errorf("metric name '%s' doesn't match regex '%s'", v, metricNameRE)
+			}
+		}
+
+		if _, ok := currentMapping.Labels["name"]; !ok {
+			return fmt.Errorf("Line %d: metric mapping didn't set a metric name", i)
```
juliusv: "Line" -> "line" (start error strings lower-cased)

mapper.go
bakins: Not sure why I did it this way. fixing.

mapper.go
bakins: FWIW, errors in `updateMapping` in this same file use uppercase "L".  I'll change my to lower.

bakins:
@juliusv Addressed your comments.

@drawks I can split this up if that's the consensus.  YAML change would need to be merged first, IMO. In https://github.com/prometheus/statsd_exporter/issues/19 it seems people did not want to mangle the current config format any more.  I guess splitting or not is up to @juliusv ?

bakins:
Added initial docs.

juliusv:
@bakins I think it's fine to just merge it as it is this time (YAML + histograms combined), although in the future I would recommend separating changes into multiple PRs.

Thanks for all the patience and sorry for being so slow here! I'll merge it and then we can still do followup cleanups as needed.

bakins:
@juliusv thanks. As with most things, this PR started very simple...

PR #70, opened 2017-05-23
Allow comments in mapping files

jasonmoo:
Hi.  This is handy and missing.  Hope it is acceptable @juliusv.


juliusv:
:+1: Thanks. Even better would be to also allow comments at the end of other lines, but it's ok for now I think.

PR #71, opened 2017-07-11
add TCP StatsD listener support

jwfang:
support TCP StatsD, on port 9125

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -437,11 +437,11 @@ samples:
 	return
 }
 
-type StatsDListener struct {
+type StatsDUDPListener struct {
 	conn *net.UDPConn
 }
 
-func (l *StatsDListener) Listen(e chan<- Events) {
+func (l *StatsDUDPListener) Listen(e chan<- Events) {
 	buf := make([]byte, 65535)
 	for {
 		n, _, err := l.conn.ReadFromUDP(buf)
@@ -452,7 +452,7 @@ func (l *StatsDListener) Listen(e chan<- Events) {
 	}
 }
 
-func (l *StatsDListener) handlePacket(packet []byte, e chan<- Events) {
+func (l *StatsDUDPListener) handlePacket(packet []byte, e chan<- Events) {
 	lines := strings.Split(string(packet), "\n")
 	events := Events{}
 	for _, line := range lines {
--- a/exporter.go
+++ b/exporter.go
@@ -484,11 +484,13 @@ func (l *StatsDTCPListener) handleConn(c *net.TCPConn, e chan<- Events) {
 		line, isPrefix, err := r.ReadLine()
 		if err != nil {
 			if err != io.EOF {
+				networkStats.WithLabelValues("tcp_error").Inc()
 				log.Errorf("Read %s failed: %v", c.RemoteAddr(), err)
 			}
 			break
 		}
 		if isPrefix {
+			networkStats.WithLabelValues("tcp_line_too_long").Inc()
 			log.Errorf("Read %s failed: line too long", c.RemoteAddr())
 			break
 		}
--- a/exporter.go
+++ b/exporter.go
@@ -14,10 +14,12 @@
 package main
 
 import (
+	"bufio"
 	"bytes"
 	"encoding/binary"
 	"fmt"
 	"hash/fnv"
+	"io"
 	"net"
 	"regexp"
 	"strconv"
@@ -289,10 +291,6 @@ func NewExporter(mapper *metricMapper, addSuffix bool) *Exporter {
 	}
 }
 
-type StatsDListener struct {
-	conn *net.UDPConn
-}
-
 func buildEvent(statType, metric string, value float64, relative bool, labels map[string]string) (Event, error) {
 	switch statType {
 	case "c":
@@ -321,17 +319,6 @@ func buildEvent(statType, metric string, value float64, relative bool, labels ma
 	}
 }
 
-func (l *StatsDListener) Listen(e chan<- Events) {
-	buf := make([]byte, 65535)
-	for {
-		n, _, err := l.conn.ReadFromUDP(buf)
-		if err != nil {
-			log.Fatal(err)
-		}
-		l.handlePacket(buf[0:n], e)
-	}
-}
-
 func parseDogStatsDTagsToLabels(component string) map[string]string {
 	labels := map[string]string{}
 	networkStats.WithLabelValues("dogstatsd_tags").Inc()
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -32,13 +32,14 @@ func init() {
...
-	readBuffer          = flag.Int("statsd.read-buffer", 0, "Size (in bytes) of the operating system's transmit read buffer associated with the UDP connection. Please make sure the kernel parameters net.core.rmem_max is set to a value greater than the value specified.")
-	addSuffix           = flag.Bool("statsd.add-suffix", true, "Add the metric type (counter/gauge/timer) as suffix to the generated Prometheus metric (NOT recommended, but set by default for backward compatibility).")
-	showVersion         = flag.Bool("version", false, "Print version information.")
+	listenAddress          = flag.String("web.listen-address", ":9102", "The address on which to expose the web interface and generated Prometheus metrics.")
+	metricsEndpoint        = flag.String("web.telemetry-path", "/metrics", "Path under which to expose metrics.")
+	statsdListenAddress    = flag.String("statsd.listen-address", ":9125", "The UDP address on which to receive statsd metric lines.")
+	statsdTCPListenAddress = flag.String("statsd.tcp-listen-address", ":9126", "The TCP address on which to receive statsd metric lines.")
```
SuperQ: I'd rather keep this one flag, so that it listens on the same TCP/UDP port.  Also it looks like you have a typo, it's listening on `9126` here.

jwfang: good catch, i fixed it.

i will make this optional.

SuperQ:
Instead of making this UDP or TCP, I would just have it listen for both.

jwfang:
sorry, misunderstand you.  

really make more sense to listen on both.

SuperQ:
No problem, thanks for working on this feature! 😃

main.go
```diff
@@ -34,7 +34,9 @@ func init() {
 var (
 	listenAddress       = flag.String("web.listen-address", ":9102", "The address on which to expose the web interface and generated Prometheus metrics.")
 	metricsEndpoint     = flag.String("web.telemetry-path", "/metrics", "Path under which to expose metrics.")
-	statsdListenAddress = flag.String("statsd.listen-address", ":9125", "The UDP address on which to receive statsd metric lines.")
+	statsdListenAddress = flag.String("statsd.listen-address", ":9125", "The UDP/TCP address on which to receive statsd metric lines.")
+	statsdListenUDP     = flag.Bool("statsd.listen-udp", true, "Whether to receive UDP statsd metrics.")
+	statsdListenTCP     = flag.Bool("statsd.listen-tcp", false, "Whether to receive TCP statsd metrics.")
```
SuperQ: This can be `true` by default.

exporter.go
```diff
@@ -351,105 +338,159 @@ func parseDogStatsDTagsToLabels(component string) map[string]string {
...
+				continue
+			}
+			events = append(events, event)
+		}
+		networkStats.WithLabelValues("legal").Inc()
+	}
+	return
+}
+
+type StatsDListener struct {
```
SuperQ: Can we make this `StatsDUDPListener` for consistency?

jwfang: i thought UDP is primary and TCP is secondary originally.

done rename.

SuperQ:
Ping @juliusv Any more comments?

exporter.go
```diff
@@ -351,105 +338,159 @@ func parseDogStatsDTagsToLabels(component string) map[string]string {
...
+	conn *net.TCPListener
+}
+
+func (l *StatsDTCPListener) Listen(e chan<- Events) {
+	for {
+		c, err := l.conn.AcceptTCP()
+		if err != nil {
+			log.Fatalf("AcceptTCP failed: %v", err)
+		}
+		go l.handleConn(c, e)
```
jacksontj: Before calling handleconn (or immediately in it) -- we should probably call `SetReadBuffer` -- to be consistent with UDP

exporter.go
```diff
@@ -351,105 +338,159 @@ func parseDogStatsDTagsToLabels(component string) map[string]string {
...
+func (l *StatsDUDPListener) handlePacket(packet []byte, e chan<- Events) {
+	lines := strings.Split(string(packet), "\n")
+	events := Events{}
+	for _, line := range lines {
+		events = append(events, lineToEvents(line)...)
 	}
 	e <- events
 }
+
+type StatsDTCPListener struct {
```
jacksontj: We don't *need* a separate listener -- assuming we have a type switch. It seems like it'd be *less* code-- but this definitely isn't wrong ;)

exporter.go
```diff
@@ -351,105 +338,159 @@ func parseDogStatsDTagsToLabels(component string) map[string]string {
...
+		go l.handleConn(c, e)
+	}
+}
+
+func (l *StatsDTCPListener) handleConn(c *net.TCPConn, e chan<- Events) {
+	defer c.Close()
+
+	r := bufio.NewReader(c)
+	for {
+		line, isPrefix, err := r.ReadLine()
```
jwfang: ReadBytes will read till *delimiter* or *eof*, so i think ReadLine is more safe for bogus input with really long line, or no line delimiter at all. :)

exporter.go
jwfang: UDP needs a 64K buffer, since it will recv all data in one packet.
but for TCP, i guess the underlying bufio is manage the buffer for us.

exporter.go
jacksontj: Fair enough, LGTM then :) Lets merge it!

SuperQ:
@juliusv care to take one last look?

exporter.go
```diff
@@ -351,105 +338,159 @@ func parseDogStatsDTagsToLabels(component string) map[string]string {
...
 }
 
-func (l *StatsDListener) handlePacket(packet []byte, e chan<- Events) {
-	lines := strings.Split(string(packet), "\n")
-	events := Events{}
-	for _, line := range lines {
-		if line == "" {
-			continue
-		}
+func lineToEvents(line string) (events Events) {
```
grobie: Please remove the named return variable here as it adds stutter to the function signature and use explicit returns below. It's not a go idiom to use naked returns in such long functions.

> Named returned values should only be used on public funcs and methods
when it contributes to the documentation.

https://go-review.googlesource.com/c/20024/

exporter.go
```diff
@@ -351,105 +338,159 @@ func parseDogStatsDTagsToLabels(component string) map[string]string {
...
+
+func (l *StatsDTCPListener) handleConn(c *net.TCPConn, e chan<- Events) {
+	defer c.Close()
+
+	r := bufio.NewReader(c)
+	for {
+		line, isPrefix, err := r.ReadLine()
+		if err != nil {
+			if err != io.EOF {
+				log.Errorf("Read %s failed: %v", c.RemoteAddr(), err)
```
grobie: It'd be cool to have metrics for these errors so that operators can set up alerts on misbehaving clients.

main.go
```diff
@@ -122,34 +138,53 @@ func main() {
 		os.Exit(0)
 	}
 
+	if !*statsdListenUDP && !*statsdListenTCP {
+		log.Fatalln("At least one of UDP/TCP listeners should be specified.")
```
grobie: Sounds like a `must` then.

main.go
```diff
@@ -34,7 +34,9 @@ func init() {
 var (
 	listenAddress       = flag.String("web.listen-address", ":9102", "The address on which to expose the web interface and generated Prometheus metrics.")
 	metricsEndpoint     = flag.String("web.telemetry-path", "/metrics", "Path under which to expose metrics.")
-	statsdListenAddress = flag.String("statsd.listen-address", ":9125", "The UDP address on which to receive statsd metric lines.")
+	statsdListenAddress = flag.String("statsd.listen-address", ":9125", "The UDP/TCP address on which to receive statsd metric lines.")
+	statsdListenUDP     = flag.Bool("statsd.listen-udp", true, "Whether to receive UDP statsd metrics.")
```
grobie: Instead of boolean flags and introducing a dependency between flags, it'd be simpler to just have to flags with the addresses for each protocol, like `-statsd.listen-address-udp=":9125"` `-statsd.listen-address-tcp=":9125"`. An empty value signals disables the server on that protocol.

main.go
```diff
@@ -122,34 +138,53 @@ func main() {
...
+	}
+
 	if *addSuffix {
 		log.Warnln("Warning: Using -statsd.add-suffix is discouraged. We recommend explicitly naming metrics appropriately in the mapping configuration.")
 	}
+
 	log.Infoln("Starting StatsD -> Prometheus Exporter", version.Info())
 	log.Infoln("Build context", version.BuildContext())
-	log.Infoln("Accepting StatsD Traffic on", *statsdListenAddress)
+	log.Infof("Accepting StatsD Traffic on %s, UDP %v, TCP %v", *statsdListenAddress, *statsdListenUDP, *statsdListenTCP)
```
grobie: That's not longer printing the address it's listening on, but `true`/`false`, but would be fixed with the comment above.

main.go
jwfang: doing it like what you mentioned have two conerns for me:
1. a bit uncomfortable for setting something to "" to disable it; and since we're default both UDP/TCP to on, use have to explicitly disable it;
2. the flags changes will break the user interface, from *statsdListenAddress* to *statsdListenUDP*. 

i have no strong feeling for 1; and for 2, if that's OK, i will do the flags changes.

jwfang:
@grobie thanks for your reivew, addressed most of your comments except for the test for TCPListener.

can't find some easy way to mock the TCPConnection. maybe we can make a 
`func (l *StatsDTCPListener) handleConn(r *io.Reader, e chan<- Events) `
 but that made the stats/logging a bit awkward. :(

grobie:
> can't find some easy way to mock the TCPConnection

I'd say, then don't mock for now and just start the server, send it a statsd line via TCP and check that it was ingested properly (by looking at the exported metrics).

jwfang:
@grobie add test for TCP listener in the same place as UDP except benchmarks.

grobie:
Thanks a lot for your contribution @jwfang!

Issue #63, opened 2017-03-08
Handling of conflicting metric values


acdha:
We had some developers on a new project starting to integrate statsd support in their application. `statsd_exporter` 0.3.0 is crashing constantly with this error message:

```
FATA[0000] A change of configuration created inconsistent metrics for "query_timer". You have to restart the statsd_exporter, and you should consider the effects on your monitoring setup. Error: a previously registered descriptor with the same fully-qualified name as Desc{fqName: "query_timer", help: "Metric autogenerated by statsd_exporter.", constLabels: {after_date="2017-01-19 00:00:00-05:00",application="…",component="…",environment="testing",server="…",sub_component="…"}, variableLabels: []} has different label names or a different help string  source=exporter.go:137
```

If I'm reading that correctly, this due to them having two versions of the app sending inconsistent metric values but while an error in this situation seems reasonable it actually causes the `statsd_exporter` process to crash.

amahomet:
guys, any news here?

grobie:
What behavior would you expect here? We can't register both conflicting metrics and export them. Just logging the error and continuing would result in silently ignoring all conflicting metrics.

acdha:
I hit this problem as a report that “Prometheus is down” when some developers were rolling out a new build with different help text. Exiting just meant that it was in a loop with Upstart restarting the service so I'm not sure that was substantially better than an ERROR log message.

grobie:
@acdha So you would prefer writing out log lines and ignoring all metrics with the new signature until someone manually restarts the statsd_exporter in that case?

acdha:
I guess it's a judgement call whether you think it's better to be disruptive, forcing people to actually notice the problem, or to allow other applications to continue sending stats without interruption.

acdha:
In my case running an instance shared by several teams, it would have been preferable if only the project which changed their stats experienced a gap but there is an argument that it'd also be acceptable to simply say “monitor process flapping better”.

grobie:
I guess I wouldn't share a statsd exporter between teams for such reasons. We also highly recommend to use direct instrumentation with client libraries. We generally tend to prefer failing hard and early, as everything else usually makes debugging very difficult. That's why I'm a bit hesitant to implement a solution which will silently ignore metrics.

In general, I'd recommend to use direct instrumentation with our client libraries instead of relying on the statsd_exporter so much.

acdha:
[...]

PR #72, opened 2017-07-18
Fixes for crashes with conflicting metric values

jacksontj:
This PR contains the fixes for #63 

I've left them as separate commits for easier review, if preferred I can squash these as well.

cc: @grobie


grobie:
Thanks!

PR #73, opened 2017-07-20
Add TCP Listener

jacksontj:
This commit adds support for the TCP protocol of statsd. In the spirit
of "be a drop-in-replacement" this TCP listener is on by default using
the same default port of 9125.

For reference, the statsd implementation -- https://github.com/etsy/statsd/pull/448

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -14,6 +14,7 @@
 package main
 
 import (
+	"bufio"
 	"bytes"
 	"encoding/binary"
 	"fmt"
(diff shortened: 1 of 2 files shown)


SuperQ:
This is already being worked on in https://github.com/prometheus/statsd_exporter/pull/71

We're just finishing up the review there.  Care to take a look?

jacksontj:
@SuperQ Done-- overall looks pretty good. The main change there I'd suggest is to swap from `ReadLine` to `ReadBytes`. Assuming the other PR merges I'll just close this out.

PR #74, opened 2017-07-20
Metricsregistry

jacksontj:
This is a WIP implementation of the feature I brought up in #63.

The goal here is to allow statsd to register a metric with multiple tag-sets -- this is important because statsd doesn't place these restrictions so if we want this to be a drop in replacement we can't have them. In addition prometheus (the server) is fine with this-- the restrictions are in place primarily as a help to application writers, since the vast majority of use-cases are people writing metrics in their app (where those restrictions are beneficial).

At this point it is all functioning, but there is a decent amount of wrapping. Most of the "wrapping" should be able to go away after the below "Upstream PRs" are closed out.

I wanted to get this up here for some initial feedback while I'm waiting on the upstream PRs.

Upstream Discussion: https://groups.google.com/forum/#!topic/prometheus-developers/Q2pRR-UlHI0

Upstream PRs I'm waiting on:

- https://github.com/prometheus/client_golang/pull/326
- https://github.com/prometheus/client_golang/pull/327

### DIFF
--- a/main.go
+++ b/main.go
@@ -34,7 +34,8 @@ func init() {
 
 var (
 	listenAddress       = flag.String("web.listen-address", ":9102", "The address on which to expose the web interface and generated Prometheus metrics.")
-	metricsEndpoint     = flag.String("web.telemetry-path", "/metrics", "Path under which to expose metrics.")
+	metricsEndpoint     = flag.String("web.telemetry-path", "/statsd", "Path under which to expose metrics.")
+	intMetricsEndpoint   = flag.String("web.int-telemetry-path", "/metrics", "Path under which to expose metrics.")
 	statsdListenAddress = flag.String("statsd.listen-address", "", "The UDP address on which to receive statsd metric lines. DEPRECATED, use statsd.listen-udp instead.")
 	statsdListenUDP     = flag.String("statsd.listen-udp", ":9125", "The UDP address on which to receive statsd metric lines. \"\" disables it.")
 	statsdListenTCP     = flag.String("statsd.listen-tcp", ":9125", "The TCP address on which to receive statsd metric lines. \"\" disables it.")
@@ -44,7 +45,7 @@ var (
 )
 
 func serveHTTP() {
-	//http.Handle(*metricsEndpoint, prometheus.Handler())
+	http.Handle(*intMetricsEndpoint, prometheus.Handler())
 	http.Handle(*metricsEndpoint, promhttp.HandlerFor(registry, promhttp.HandlerOpts{}))
 	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
 		w.Write([]byte(`<html>
--- a/main.go
+++ b/main.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/howeyc/fsnotify"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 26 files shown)


jacksontj:
After conversations with upstream, they don't want to change the upstream client -- so we'll just have to keep track of the wrapping in this PR (similar to how it exists). I'll be refactoring most of the code  out into a package in this repository (since it could be replaced in the future -- and might be useful elsewhere).

lightsaway:
Will this be merged?

jacksontj:
@lightsaway I need to do some code-cleanup first, but then it should be. I plan on getting to it this week.

lightsaway:
@jacksontj Thanks!

drawks:
Any progress here? I'm eagerly awaiting some improved flexibility of this exporter.

jacksontj:
Apologies for the delays, this fell down on my priority list-- but I do plan on cleaning this up and getting it merged (going to be more work as there are now conflicts). My plan is to still get this merged in at some point.

matthiasr:
Would this also address #80 by no longer including the exporter's own metrics?

Or even better, would it allow putting these on a different endpoint?

jacksontj:
@matthiasr that was an oustanding question actually-- from the discussion on #80 it seems like a split is the way to go-- with that I've just added ca1f17b which splits them into 2 endpoints.

jacksontj:
Since none of the PRs are going to be merged, this should be good to merge. There's the added question of what to do with the 2 registries. For now I have them split, but this *is* a backwards incompatible change (i'm not sure what the rules are around releases for this project).

main.go
```diff
@@ -34,7 +34,8 @@ func init() {
 
 var (
 	listenAddress       = flag.String("web.listen-address", ":9102", "The address on which to expose the web interface and generated Prometheus metrics.")
-	metricsEndpoint     = flag.String("web.telemetry-path", "/metrics", "Path under which to expose metrics.")
+	metricsEndpoint     = flag.String("web.telemetry-path", "/statsd", "Path under which to expose metrics.")
+	intMetricsEndpoint   = flag.String("web.int-telemetry-path", "/metrics", "Path under which to expose metrics.")
```
matthiasr: Please update the help string for both flags to make the difference very very clear.

registry.go
```diff
@@ -0,0 +1,328 @@
...
+
+	"github.com/golang/protobuf/proto"
+	"github.com/prometheus/client_golang/prometheus"
+
+	dto "github.com/prometheus/client_model/go"
+)
+
+const (
+	// Capacity for the channel to collect metrics and descriptors.
+	capMetricChan = 1000
```
matthiasr: What happens when these are exceeded? What happens for someone with very many metrics?

registry.go
```diff
@@ -0,0 +1,328 @@
...
+)
+
+// For statsd all our metrics will be both a Metric and a collector -- to make life easier
+// we'll define an interface
+type CollectorMetric interface {
+	prometheus.Metric
+	prometheus.Collector
+}
+
+// TODO: remove once https://github.com/prometheus/client_golang/pull/326 is merged
```
matthiasr: that PR is closed. what's the future here?

matthiasr:
I'm closing this because it is very stale and inactive. If anyone wants to pick up the inconsistent label problem, I think the correct way is to solve this on our end instead of forking the client library.

PR #79, opened 2017-08-01
Resolves #77 add cmdline flag to drop unmapped

### LINKED ISSUE #77: Allow for dropping unmapped metrics
The default behavior of passing unmapped metrics through can result in really polluted metric namespaces when using software that generates a large number of dynamic metrics OR a large number of metrics that are irrelevent to the user's interest. I'd like to be able to switch the behavior in the config to blackholing unmapped metrics.

drawks:
* Adds `-statsd.drop-unmapped` cmdline flag which causes statsd metrics
  that don't match a mapping to be dropped. New flag defaults to "false"

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -266,6 +266,9 @@ func (b *Exporter) Listen(e <-chan Events) {
 				}
 			} else {
 				eventsUnmapped.Inc()
+				if *dropUnmapped == true {
+					continue
+				}
 				metricName = escapeMetricName(event.MetricName())
 			}
 
--- a/exporter.go
+++ b/exporter.go
@@ -218,12 +218,13 @@ func (c *TimerEvent) Labels() map[string]string { return c.labels }
 type Events []Event
 
 type Exporter struct {
-	Counters   *CounterContainer
-	Gauges     *GaugeContainer
-	Summaries  *SummaryContainer
-	Histograms *HistogramContainer
-	mapper     *metricMapper
-	addSuffix  bool
+	Counters     *CounterContainer
+	Gauges       *GaugeContainer
+	Summaries    *SummaryContainer
+	Histograms   *HistogramContainer
+	mapper       *metricMapper
+	addSuffix    bool
+	dropUnmapped bool
 }
 
 func escapeMetricName(metricName string) string {
@@ -266,7 +267,7 @@ func (b *Exporter) Listen(e <-chan Events) {
 				}
 			} else {
 				eventsUnmapped.Inc()
-				if *dropUnmapped == true {
+				if b.dropUnmapped == true {
 					continue
 				}
 				metricName = escapeMetricName(event.MetricName())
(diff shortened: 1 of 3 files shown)


SuperQ:
I think it would be worth adding a test for this condition to `exporter_test.go`.

drawks:
I've been staring at this code this morning and I don't see any straightforward way to test this, of course `exporter_test.go` is relatively bare indicating that perhaps others have also found this code hard to write tests against.

drawks:
@SuperQ I've added a test as requested.

exporter.go
```diff
@@ -266,6 +266,9 @@ func (b *Exporter) Listen(e <-chan Events) {
 				}
 			} else {
 				eventsUnmapped.Inc()
+				if *dropUnmapped == true {
```
grobie: It's unnecessary and uncommon in Golang to compare a boolean value with a boolean value. Please simplify to `if *dropUnmapped {`.

main.go
```diff
@@ -41,6 +41,7 @@ var (
 	readBuffer          = flag.Int("statsd.read-buffer", 0, "Size (in bytes) of the operating system's transmit read buffer associated with the UDP connection. Please make sure the kernel parameters net.core.rmem_max is set to a value greater than the value specified.")
 	addSuffix           = flag.Bool("statsd.add-suffix", true, "Add the metric type (counter/gauge/timer) as suffix to the generated Prometheus metric (NOT recommended, but set by default for backward compatibility).")
 	showVersion         = flag.Bool("version", false, "Print version information.")
+	dropUnmapped        = flag.Bool("statsd.drop-unmapped", false, "Drop statsd metrics which have no matched mapping configured.")
```
grobie: Let's group at least all flags with the same namespace (e.g. `statsd` together).

exporter.go
grobie: Uh, please don't rely on a global variable. This setting should be an attribute of the exporter itself and configured when initializing the exporter.

exporter_test.go
```diff
@@ -53,6 +53,45 @@ func TestNegativeCounter(t *testing.T) {
...
+	go func() {
+		time.Sleep(time.Millisecond * 100)
+		close(events)
+	}()
+
+	ex.Listen(events)
+	if len(ex.Counters.Elements) != 0 {
+		t.Fatalf("Unmapped metric recorded inspite of dropUnmapped being set true")
+	}
+	dropUnmapped = func() *bool { b := false; return &b }()
```
grobie: The fact that you have to jump through such hoops and introduce side effects in a test function by changing the behavior of a global variable is a strong indicator to avoid global variables altogether.

drawks:
@grobie please see updated branch that addresses your line review.

As for it being a cmd line option vs a property of the mapping rules. I was just following the existing pattern of `addSuffix` and wanted to allow this functionality regardless of whether using the new yaml mapping configs or the legacy mapping configs. I really dislike the legacy mapping configs and didn't want to monkey with that code path, but also didn't want my feature to be ONLY available to users of the new yaml syntax.

grobie:
I've thought about it for a bit, and I feel quite strongly about not adding more command line flags to change the behavior of the collector. I'm also against dragging old code around for long. The transition to yml configs is easy, old binaries will still be around for users if they have to use the old format for whatever reason.

I'd vote for the following changes for the next release:
- remove old config format
- remove the flag to automatically append the wrong _counter / etc. suffixes
- add an action: attribute which defaults to `map` or something.
- add your addition here as `action: drop`

WDYT?

drawks:
Sounds great to me, my own dev branch already has dropped all support for legacy config format. The other changes you've mentioned shouldn't be too complicated to implement. I've got a bunch of $dayjob work this week, but I'll see if I can sneak some time into it.

drawks:
Yeah my branch has a a config key for `action` that can either `map` or `drop` and defaults to `map`. I also added a new `match_type` of `all` so we don't need to eat the performance of regex to have a default action.

matthiasr:
Could you push those changes?

matthiasr:
The drop action was added in #115.

PR #85, opened 2017-08-03
Resolves #84 adds support for regex mappers

### LINKED ISSUE #84: Support raw regex for mapping in addition to globbing
While trying to use this exporter with some particularly heinous 3rd party software I quickly ran into the limitation that I would like to be able to define mappings which are more precisely defined than what is allowed in the current globbing match syntax. Looking through the code, the globbing is actually handled internally by being translated to regex.

I'd propose that there should be a configuration option available utilizing the newer yaml mapper configuration that allows for selection between `match_type` with a value of either `glob` or `regex`.

I've been testing the changes necessary to implement this and will have a PR shortly.

drawks:
* Added `regex` match_type as valid config callout in both individual mappers
    and mapper defaults
  * Added test coverage for changes
  * Updated documentation to reflect usage of `match_type` parameter and
    provide example of a raw regex match rule

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,23 @@ mappings:
     job: "${1}_server"
 ```
 
+Another capability when using YAML configuration is the ability to define matches
+using raw regular expressions as opposed to the default globbing style of match.
+This may allow for pulling structured data from otherwise poorly named statsd
+metrics AND allow for more precise targetting of match rules. When no `match_type`
+paramter is specified the default value of `glob` will be assumed:
+```yaml
+mappings:
+- match: (.*)\.(.*)--(.*)\.status\.(.*)\.count
+  match_type: regex
+  labels:
+    name: "request_total"
+    hostname: "$1"
+    exec: "$2"
+    protocol: "$3"
+    code: "$4"
+```
+
 Note, that one may also set the histogram buckets.  If not set, then the default
 [Prometheus client values](https://godoc.org/github.com/prometheus/client_golang/prometheus#pkg-variables) are used: `[.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10]`. `+Inf` is added
 automatically.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -169,6 +169,23 @@ mappings:
     job: "${1}_server"
 ```
 
+Another capability when using YAML configuration is the ability to define matches
+using raw regular expressions as opposed to the default globbing style of match.
+This may allow for pulling structured data from otherwise poorly named statsd
+metrics AND allow for more precise targetting of match rules. When no `match_type`
+paramter is specified the default value of `glob` will be assumed:
+```yaml
```
grobie: Please keep a newline between text paragraphs and text blocks.

match.go
```diff
@@ -0,0 +1,41 @@
...
+		return err
+	}
+
+	switch matchType(v) {
+	case matchTypeRegex:
+		*t = matchTypeRegex
+	case matchTypeGlob, matchTypeDefault:
+		*t = matchTypeGlob
+	default:
+		return fmt.Errorf("invalid match type '%s'", v)
```
grobie: By the way, you can use `%q` to avoid manual quoting of strings around `%s`.

mapper.go
```diff
@@ -156,12 +158,22 @@ func (m *metricMapper) initFromYAMLString(fileContents string) error {
...
+		if currentMapping.MatchType == "" {
+			currentMapping.MatchType = n.Defaults.MatchType
+		}
 
-		// Translate the glob-style metric match line into a proper regex that we
-		// can use to match metrics later on.
-		metricRe := strings.Replace(currentMapping.Match, ".", "\\.", -1)
-		metricRe = strings.Replace(metricRe, "*", "([^.]+)", -1)
-		currentMapping.regex = regexp.MustCompile("^" + metricRe + "$")
+		if currentMapping.MatchType == matchTypeGlob {
```
grobie: I'd have used a case statement, switching on the value of `currentMapping.MatchType` and an explicit default case raising an error on an unknown matchtype.

drawks:
@grobie I touched up the PR to get 2 of your 3 suggestions.

matt-martin:
This would be very useful!

drawks:
any feedback here, I got an approval two weeks ago, but no merge and no further review.

SuperQ:
I agree with @grobie's suggestion to use a case statement.  But I don't feel strongly about it. Otherwise LGTM.

grobie:
@drawks all your PRs are on my queue, which is quite long. I'm really tight on time at the moment, and there are some prometheus 2.0 projects which need priority at the moment. I'm sorry for the delays, maybe others can help in the meantime.

PR #87, opened 2017-08-03
Resolves #78 Allow help text to be customized

### LINKED ISSUE #78: Allow for setting "#help" text in mapping config
The current behavior of providing a help text like 
```# HELP test_case_total Metric autogenerated by statsd_exporter.```

falls pretty far short of the intent of informing the consumer of the metric what the metric represents in easy language. This should be a very small modification that would improve usability.

### PR DESCRIPTION
**Pull request by drawks**
* Updated mappings to accept custom help text in YAML config
  * Updated exporter to display configured help messages
  * Update README to reflect aditional configurability
  * Minor Update to exporter to allow tests to pass with mocked
    statsdlistener that has no mappers

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -71,13 +71,19 @@ func NewCounterContainer() *CounterContainer {
 	}
 }
 
-func (c *CounterContainer) Get(metricName string, labels prometheus.Labels) (prometheus.Counter, error) {
+func (c *CounterContainer) Get(metricName string, labels prometheus.Labels, mapping *metricMapping) (prometheus.Counter, error) {
+	var help string
+	if mapping.HelpText == "" {
+		help = defaultHelp
+	} else {
+		help = mapping.HelpText
+	}
 	hash := hashNameAndLabels(metricName, labels)
 	counter, ok := c.Elements[hash]
 	if !ok {
 		counter = prometheus.NewCounter(prometheus.CounterOpts{
 			Name:        metricName,
-			Help:        defaultHelp,
+			Help:        help,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by juliusv** [comment]
Sorry for the wait!

Seems like this could be much shorter and more targeted if we didn't pass the mapping into the individual getter methods, but just the help (since that's the only thing that's used from it)? Then we also need the code to set it once.

File: exporter.go [hunk 1]
```diff
@@ -98,13 +104,19 @@ func NewGaugeContainer() *GaugeContainer {
 	}
 }
 
-func (c *GaugeContainer) Get(metricName string, labels prometheus.Labels) (prometheus.Gauge, error) {
+func (c *GaugeContainer) Get(metricName string, labels prometheus.Labels, mapping *metricMapping) (prometheus.Gauge, error) {
+	var help string
```
grobie: Instead of copy&pasting the setting of the helptext here 4 times, I'd just ensure that the mapping is initialized correctly. Maybe by introducing a `newMapping()` function which sets help to defaultHelp.

mapper.go
```diff
@@ -52,8 +52,16 @@ type metricMapping struct {
 	TimerType timerType         `yaml:"timer_type"`
 	Buckets   []float64         `yaml:"buckets"`
 	MatchType matchType         `yaml:"match_type"`
+	HelpText  string            `yaml:"help"`
 }
 
+type configLoadStates int
```
grobie: What's this? Accidental rebase leftovers?

drawks: yah I think that's it, lemme clean it up

PR #89, opened 2017-08-04
Resolves #20 Allow for templated metric names

### LINKED ISSUE #20: Mapped metric name cannot be dynamic
When I tried taking parts of the parsed output and use it for the metric name, for example: "name": "$2_$3", it seems that statsd_bridge fail to start if I configure it and complains about it.
Using a static name it works.

drawks:
* Removed field validation for "name" label
  * Updated tests to match
  * Wraped metric names in escape logic in exporter

I suppose this should probably have a bit of a discussion. Due to the way the code works enabling templating in in the name label of a mapping only required removing a regex validator that checked that the name label was a valid prometheus metric name. Since you can't validate a template until it is expanded it makes sense that we wouldn't have the additionally validation on this parameter. So as an alternative I've just wrapped the name of even mapped metrics with `escapeMetricName` to ensure the final name is valid.

Additionally, there is the caveat that once the full metric definition and labels are dynamic there are myriad of way to get unanticipated metric namespace collisions. For instance it is possible to end up with the same metric with two different label sets which the prometheus exporter library won't allow. There was some effort in #74 to replace the stock registry from the prometheus go exporter with a more permissive registry; this would probably make sense (IMHO) as it would definitely allow for an easier migration from statsd to prometheus. However the registry code in #74 is still quite rough.

TLDR; this allows dynamic metric names, flexibility isn't free, shooting yourself in the foot becomes easier.

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -258,7 +258,7 @@ func (b *Exporter) Listen(e <-chan Events) {
 
 			mapping, labels, present := b.mapper.getMapping(event.MetricName())
 			if present {
-				metricName = labels["name"]
+				metricName = escapeMetricName(labels["name"])
 				for label, value := range labels {
 					if label != "name" {
 						prometheusLabels[label] = value
--- a/mapper.go
+++ b/mapper.go
@@ -144,12 +144,9 @@ func (m *metricMapper) initFromYAMLString(fileContents string) error {
 		}
 
 		// Check that label is correct.
-		for k, v := range currentMapping.Labels {
+		for k := range currentMapping.Labels {
 			if !metricNameRE.MatchString(k) {
 				return fmt.Errorf("invalid label key: %s", k)
-			}
-			if k == "name" && !metricNameRE.MatchString(v) {
-				return fmt.Errorf("metric name '%s' doesn't match regex '%s'", v, metricNameRE)
 			}
 		}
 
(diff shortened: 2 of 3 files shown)


drawks:
@grobie any chance I can get some feedback on this?

matthiasr:
@drawks do you have time to work on this? The change itself is good, it will conflict with #101 but not fundamentally; I would like some positive test cases too. Also please document how to use this in the README.

drawks:
I have a branch will all of those things on my workstation, but have been super busy with $dayjob. Actually I have the same function as #101 in my branch as well

matthiasr:
Great! Can you push them?

Let's do one change at a time, but feel free to either rebase straight onto #101 or rebase onto master and then we can fix it up later. I can also help getting things into shape then.

drawks:
The biggest time sink is a general refactor of the mapper/config parsing code to enable better tests. I can't commit to a timeline to get this completed unfortunately as $dayjob has shifted priorities a bit.

matthiasr:
Is this possible to _reasonably_ test without a major refactoring? For example, just by adding scenarios to `mapper_test.go` to test it through the whole component?

I think this is a really useful feature and would like to have it in.

matthiasr:
This is implemented via #117 now!

PR #92, opened 2017-08-14
Resolves #81 - Reduce log levels

### LINKED ISSUE #81: Malformed statsd lines should be reported at lower log level.
```ERRO[0785] Error building event on line docker-lb2.fabio--garbage.tacostand_com./.12_117_14_4_21506.mean:79292673.35|gf: Bad stat type gf  source=exporter.go:51```

Is an example of a badly formed statsd metric for the non-existent type `gf` which raises an `error` level log message from statsd_exporter. This doesn't smell like an error and should probably be downgraded to `info` level. 

https://dave.cheney.net/2015/11/05/lets-talk-about-logging

### PR DESCRIPTION
**Pull request by drawks**
* Reduced all `log.Error*` events to `log.Debug*` in `exporter.go` where
  an existing internal metric tracks the event.

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -275,6 +275,7 @@ func (b *Exporter) Listen(e <-chan Events) {
 				// will cause the exporter to panic. Instead we will warn and continue to the next event.
 				if event.Value() < 0.0 {
 					log.Debugf("Counter %q is: '%f' (counter must be non-negative value)", metricName, event.Value())
+					eventStats.WithLabelValues("illegal_negative_counter").Inc()
 					continue
 				}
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: exporter.go [hunk 1]
```diff
@@ -274,7 +274,7 @@ func (b *Exporter) Listen(e <-chan Events) {
 				// We don't accept negative values for counters. Incrementing the counter with a negative number
 				// will cause the exporter to panic. Instead we will warn and continue to the next event.
 				if event.Value() < 0.0 {
-					log.Errorf("Counter %q is: '%f' (counter must be non-negative value)", metricName, event.Value())
+					log.Debugf("Counter %q is: '%f' (counter must be non-negative value)", metricName, event.Value())
```
SuperQ: This one doesn't seem to have a metric associated with it.  But, I don't think it's that important.

grobie: Let's add such a metric/error counter before reducing the log level though. Silently dropping metrics would be very annoying.

drawks:
@grobie See updated PR branch, added internal metric as requested

PR #101, opened 2017-10-30
Allow using name in mapping rather than 'name' label

### LINKED ISSUE #90: metric name should be split from labels in yaml mapping format
#66 was recently merged and I'm a huge fan of the new yaml mappings, however I think it is probably a bad idea to have the `name` label have magic behavior in setting the metric name in the exporter. I /think/ it probably makes more sense to have a `metric_name` parameter one level higher in the mapper config. This should disambiguate the result of the mapping AND allow for, if desired, a mapping which sets a label called `name`.

Proposed change...
this:
```
mappings:
- match: test.case.*
  labels:
    name: "test_case_total"
    blargle: "$1"
```
Would become this:
```
mappings:
- match: test.case.*
  metric_name: "test_case_total"
  labels:
    blargle: "$1"
```

bakins:
fixes #90

I had to modify the tests a bit for this to work as expected. I'm not 100% that the tests are correct.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,15 +101,15 @@ An example mapping configuration:
 ```yaml
 mappings:
 - match: test.dispatcher.*.*.*
+  name: "dispatcher_events_total"
   labels:
-    name: "dispatcher_events_total"
     processor: "$1"
     action: "$2"
     outcome: "$3"
     job: "test_dispatcher"
-- match: "*.signup.*.*"
+- match: *.signup.*.*
+  name: "signup_events_total"
   labels:
-    name: "signup_events_total"
     provider: "$2"
     outcome: "$3"
     job: "${1}_server"
@@ -127,6 +127,7 @@ follows:
     test.web-server.foo.bar
      => test_web_server_foo_bar{}
 
+Each mapping in the configuration file must define a `name` for the metric.
 
 If the default metric help text is insufficient for your needs you may use the YAML
 configuration to specify a custom help text for each mapping:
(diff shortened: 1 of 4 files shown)


matthiasr:
I just realized that there has never been a release with the YAML format. In that case, I would rather remove the special handling of the "name" label altogether.

bakins:
Sgtm. I can remove the name label special handling and update the readme. That sound okay?

matthiasr:
That would be wonderful!

README.md
```diff
@@ -127,15 +127,16 @@ follows:
     test.web-server.foo.bar
      => test_web_server_foo_bar{}
 
+Each mapping in the configutaiton file must define a `name` for the metric.
```
matthiasr: typo: configutation

matthiasr:
Also still needs a rebase, apparently. Please rebase against current master instead of merging master into this branch, I find the resulting history more useful.

bakins:
@matthiasr this wasn't ready to push again - I pushed too soon - and now my branch is in an interesting state. I will have to revisit later when I have time to work through conflicts and merges.

matthiasr:
Alles klar! Let me know if you need help. FWIW – I would squash the whole thing into a single commit, since it's a single well-delineated change. That should also be easier than trying to restore individual commits.

bakins:
@matthiasr I unexpectedly had a few minutes so I have a single commit now based on current master. Thanks!

matthiasr:
Thank you!

PR #106, opened 2017-11-10
Break out the telemetry during sample processing

matthiasr:
The "packets" metric had heavily overloaded meaning for different
"outcomes", and would often be incremented multiple times, sometimes
even with a single (per-line) increment in one outcome corresponding to
multiple increments in another.

This removes the broken metric, and replaces it with separate total and
error counters for each level of processing. This allows monitoring the
network traffic handled separately from the samples incurred by it.

Supersedes #54 based on the discussion there.

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -404,14 +404,14 @@ func buildEvent(statType, metric string, value float64, relative bool, labels ma
 
 func parseDogStatsDTagsToLabels(component string) map[string]string {
 	labels := map[string]string{}
-	networkStats.WithLabelValues("dogstatsd_tags").Inc()
+	tagsReceived.Inc()
 	tags := strings.Split(component, ",")
 	for _, t := range tags {
 		t = strings.TrimPrefix(t, "#")
 		kv := strings.SplitN(t, ":", 2)
 
 		if len(kv) < 2 || len(kv[1]) == 0 {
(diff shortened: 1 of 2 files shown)


matthiasr:
@bakins WDYT about getting this into the 0.5.0 release? I'd either merge it now & then you can rebase the release branch, or I hold it back for the next release. it's technically another breaking chagne, and I kinda prefer bunching them up, OTOH the old metric was so bad I'm not too worried about breaking it.

bakins:
@matthiasr is fine with me to include in 0.5.0

exporter.go
```diff
@@ -562,21 +564,24 @@ func (l *StatsDTCPListener) Listen(e chan<- Events) {
...
 			if err != io.EOF {
-				networkStats.WithLabelValues("tcp_error").Inc()
+				tcpErrors.Inc()
 				log.Debugf("Read %s failed: %v", c.RemoteAddr(), err)
 			}
 			break
 		}
 		if isPrefix {
-			networkStats.WithLabelValues("tcp_line_too_long").Inc()
+			tcpLineTooLong.Inc()
```
grobie: Uhm, we should just fix the reader though, no need to abort on a long line. https://golang.org/pkg/bufio/#Reader.ReadLine

matthiasr:
Yes, this should be really easy to fix, but this PR is about the metrics.

Issue #102, opened 2017-11-02
"missing terminating newline" error when creates config in k8s


eskornev:
Hello!
There is situation when statsd_exporter won't start if there is no newline in config file.
We are using kubernetes 1.6 and unfortunately it creates config files from configmaps 
with no newline at the end of file... Is it possible to eliminate this restriction?

eskornev:
Just builded version 0.4.0 from source and stumbled upon yaml config file - it seems like statd exporter dont like * in the beginning of mapping match string. This config fails:

mappings:
- match: *.signup.*.*
  labels:
    name: "signup_events_total"
    provider: "$2"
    outcome: "$3"
    job: "${1}_server"

Error is: FATA[0000] Error loading config:yaml: line 1: did not find expected alphabetic or numeric character  source="main.go:193"

But this config is ok:
mappings:
- match: abcabc.*.signup.*.*
  labels:
    name: "signup_events_total"
    provider: "$2"
    outcome: "$3"
    job: "${1}_server"

matthiasr:
I think this is a limitation of the YAML format – it tries to parse `*something` as an alias to some anchor defined elsewhere, but `.signup` is not a valid anchor either. To fix this, you need to enclose the problematic string in quotes, like

```yaml
mappings:
- match: "*.signup.*.*"
  labels:
    name: "signup_events_total"
    provider: "$2"
    outcome: "$3"
    job: "${1}_server"
```

matthiasr:
Oh, I think there is the problem:

> builded version 0.4.0 from source and stumbled upon yaml config file

0.4.0 does not support the YAML config yet. I'm sorry about the confusion, basically everything about the configuration format has changed since the last release, so the README you see on the project page is not very helpful if you are using 0.4.0.

There will be a release soon (#100), until then please try building from master. Keep in mind that the config format will change once more (#101) but this will be very easy to adapt to.

Issue #97, opened 2017-09-25
Documentation and real behaviour of metric name mapping not match


nordicdyno:
README contains:

> Metrics that don't match any mapping in the configuration file are translated into Prometheus metrics without any labels and with certain characters escaped (_ -> __; - -> __; . -> _)

But in reality metrics with names `metric_name1`, `metric-name2`, `metric.name3` become `metric_name1`, `metric_name2`, `metric_name3` accordingly.

there is the code to reproduce this doc vs reality mismatch https://gist.github.com/nordicdyno/061a50dc29336dc6a60d4d0470edd151

matthiasr:
Is the current real behavior acceptable?

matthiasr:
It seems this has been the case since 728bdc52ae71d9cb9bcbda7d93f47689e94eac4c. I'm not inclined to change the behaviour, I'll fix the documentation instead.

PR #107, opened 2017-11-10
Fix and test examples in the README

matthiasr:
Following up on #102 and #97 sent me down a rabbit hole of trying to understand what actually is the expected behaviour. This PR adds tests to confirm the current behaviour, and adjusts the documentation to actually match that.

### DIFF
--- a/exporter_test.go
+++ b/exporter_test.go
@@ -153,3 +153,21 @@ func (ml *mockStatsDTCPListener) handlePacket(packet []byte, e chan<- Events) {
 	}
 	ml.handleConn(sc, e)
 }
+
+func TestEscapeMetricName(t *testing.T) {
+	scenarios := map[string]string{
+		"clean":                   "clean",
+		"0starts_with_digit":      "_0starts_with_digit",
+		"with_underscore":         "with_underscore",
+		"with.dot":                "with_dot",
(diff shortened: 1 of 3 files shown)


exporter_test.go
```diff
@@ -153,3 +153,21 @@ func (ml *mockStatsDTCPListener) handlePacket(packet []byte, e chan<- Events) {
...
+		"0starts_with_digit":      "_0starts_with_digit",
+		"with_underscore":         "with_underscore",
+		"with.dot":                "with_dot",
+		"with😱emoji":              "with_emoji",
+		"with.*.multiple":         "with___multiple",
+		"test.web-server.foo.bar": "test_web_server_foo_bar",
+	}
+
+	for in, want := range scenarios {
+		if got := escapeMetricName(in); want != got {
```
grobie: You could also just reuse Prometheus' strutil.SanitzeLabelName.

matthiasr: good point, but that doesn't do the prefixing if the first character is a digit. should it?

I don't want to change behavior at this point, so I'll stick with the function as it is right now.

grobie: Actually, it's not a guide idea to use it for this use case. It's not made to sanitize metric names, and not worth adding a dependency. Nevermind.

PR #117, opened 2018-01-16
Allow for dynamic metric name

erickpintor:
Hi folks,

I'm opening this PR as an attempt to get this feature merged since I need it in one of my projects. It was originally implemented by @drawks at the PR #89. I've reimplemented the same feature on top of the most recent version of the master branch and added a few test cases.

I hope this helps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,18 @@ mappings:
     provider: "$1"
 ```
 
+The metric name can also contain references to regex matches. The mapping above
+could be written as:
+
+```
+mappings:
+- match: test\.(\w+)\.(\w+)\.counter
+  match_type: regex
+  name: "${2}_counter"
+  labels:
+    provider: "$1"
+```
+
 If the default metric help text is insufficient for your needs you may use the YAML
 configuration to specify a custom help text for each mapping:
 ```yaml
--- a/README.md
+++ b/README.md
@@ -147,6 +147,9 @@ mappings:
     provider: "$1"
 ```
 
+Please note that metrics with the same name must also have the same set of
+labels names.
+
 If the default metric help text is insufficient for your needs you may use the YAML
 configuration to specify a custom help text for each mapping:
 ```yaml
--- a/README.md
+++ b/README.md
@@ -123,7 +123,17 @@ follows:
     test.web-server.foo.bar
      => test_web_server_foo_bar{}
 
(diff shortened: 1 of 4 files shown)


drawks:
documentation needs a bit of caveat about colliding names; i.e. each metric can only have one unique set of labels.

README.md
```diff
@@ -123,7 +123,17 @@ follows:
     test.web-server.foo.bar
      => test_web_server_foo_bar{}
 
-Each mapping in the configuration file must define a `name` for the metric.
+Each mapping in the configuration file must define a `name` for the metric. The
+metric's name can contain `$n`-style references to be replaced by the n-th
+wildcard match in the matching line. That allows for dynamic rewrites, such as:
```
matthiasr: Can you also mention how this works with the regex match type?

erickpintor:
Sure. Could you please clarify what happens when the metric names collide? I just want to make sure the docs are accurate.

matthiasr:
If two metrics have the same name, then they also must have the same label
set (set of label names, they can have different values of course).
Otherwise the /metrics endpoint will throw errors. That's already the case
but the risk is higher that it might happen accidentally with templating.

erickpintor:
@matthiasr could you take another pass. I'm not sure if the docs are clear enough.

README.md
```diff
@@ -123,7 +123,32 @@ follows:
...
+mappings:
+- match: test\.(\w+)\.(\w+)\.counter
+  match_type: regex
+  name: "${2}_counter"
+  labels:
+    provider: "$1"
+```
+
+Please note that metrics with the same name must also have the same set of
+labels names.
```
matthiasr: there's a mini typo here: "labels names" – I can fix that up myself though.

erickpintor:
By the way, I was using the docker image before finding out I needed this feature. Do you know by any chance when we can get a new version of the docker image?

matthiasr:
It should be available at `prom/statsd-exporter:master` in a few minutes. I would appreciate if you could test it under real-life conditions before I make a release.

matthiasr:
(this doesn't need to be a very extended test, if it works for you I'd make a release tomorrow or so)

erickpintor:
Sure. I can test the master tag.
I'm already using that code in a docker container, I'm just using my fork instead of the official image.

matthiasr:
Alright then, release time!

PR #119, opened 2018-01-27
Create one metric per metric

matthiasr:
This is the rough cut of a stab at addressing the many-label-values problem in #76.

To quantify the impact, I added a benchmark on `prometheus.Gather()` after feeding it with varying numbers of metrics. This could certainly be done more elegantly – for example, instead of using all the exporter machinery in Listen we could just observe some metric values directly.

Along the way, I upgraded client_golang.

I profiled specific cases of the benchmark after this change. The 1M metric names / 1 label value variant is mostly equivalent to the old situation – loads of individual metrics objects. The memory profile shows 55% of memory being allocated for new goroutines. The 1 metric name / 1M label values variant is the most extreme case of the improved behaviour for the common case of low cardinalities of label names. Goroutines don't even show up in the memory profile.

These were generated with

```
for variant in m_1_l_1000000 m_1000000_l_1; ./statsd_exporter.test -test.bench BenchmarkGather/$variant -test.cpuprofile $variant.cpu.pb.gz -test.memprofile $variant.mem.pb.gz; end
```

The full benchmark run went like this:

```
=== RUN   TestHandlePacket
--- PASS: TestHandlePacket (0.01s)
=== RUN   TestNegativeCounter
--- PASS: TestNegativeCounter (0.10s)
=== RUN   TestInvalidUtf8InDatadogTagValue
--- PASS: TestInvalidUtf8InDatadogTagValue (0.20s)
=== RUN   TestHistogramUnits
--- PASS: TestHistogramUnits (0.10s)
=== RUN   TestEscapeMetricName
--- PASS: TestEscapeMetricName (0.00s)
=== RUN   TestMetricMapperYAML
--- PASS: TestMetricMapperYAML (0.00s)
=== RUN   TestAction
--- PASS: TestAction (0.00s)
goos: darwin
goarch: amd64
pkg: github.com/prometheus/statsd_exporter
BenchmarkExporter1-4    	   50000	     23917 ns/op	    6976 B/op	     169 allocs/op
BenchmarkExporter5-4    	    3000	    429068 ns/op	  172386 B/op	    4177 allocs/op
BenchmarkExporter50-4   	      30	  50439746 ns/op	17205779 B/op	  417552 allocs/op
BenchmarkGenerator/m_10_l_10-4         	   50000	     36641 ns/op	   10490 B/op	     263 allocs/op
BenchmarkGenerator/m_100_l_0-4         	   50000	     25205 ns/op	    3846 B/op	     199 allocs/op
[...]

Issue #76, opened 2017-08-01
Upgrade client_golang to reduce memory usage with many metrics


jondot:
Hi, 
I'm seeing a ratio of 10k metrics per 100M of RAM, for production metrics. Interested to hear first if that's normal, and then, what's a good strategy for handling 1M metrics (this also affects the `metrics` endpoint polled by prom, obviously).

In my tests for 1M metrics, I've gotten 28GB of RAM and growing, and all cores max out (had to kill the process).

SuperQ:
Are you running one central `statd_exporter`?

The general recommendation is to run one exporter per node, and configure all apps to send to `localhost`.  This allows for easy scaling and eliminates the exporter as a SPoF.

It also helps if you are using UDP to transport statsd metrics, as packet loss is less of an issue.

matthiasr:
That comes out as ~10KiB per metric, that is indeed a lot.

As @superq says, the first thing I would do is colocate the exporter with every data source, instead of the central statsd server.

I am still interested in getting to the bottom of the memory usage. What are all these bytes? Do we really need them? Any input is welcome!

shuz:
We hit the same issue. We find that it can be realted to the summary metric type for timers, this is very expensive in both CPU and memory if you have high rate API calls. The most simple solution is switching to histogram.

If it's still necessary to use summary, try tune the MaxAge, BufCap and AgeBuckets. MaxAge is the key among these parameters.

matthiasr:
I got a notification for a comment that made good points but appears to be lost?

This might be related to summaries with quantile – if you have a lot of timer metrics, these are translated into summaries and can be this expensive.

Now, generally histograms are probably a better idea here, but open the question of how to configure buckets. On the other hand, one way to mitigate the memory cost of quantiles would be to make the MaxAge, BufCap and AgeBuckets also configurable, so the complexity would be the same.

How should this look in configuration?

shuz:
@matthiasr I deleted the comment because I find my metrics is still OOM after I switch to histogram (default buckets). And the calculated average memory is still quite high. You can try it using this configuration:
```
defaults:
  timer_type: histogram
```

shuz:
MaxAge, BufCap and AgeBuckets are not configurable now. 

I'm trying to make them configurable globally in my branch:
```
func (c *SummaryContainer) Get(metricName string, labels prometheus.Labels, help string) (prometheus.Summary, error) {
	hash := hashNameAndLabels(metricName, labels)
[...]

PR #120, opened 2018-01-27
Metricsregistry

### LINKED ISSUE #63: Handling of conflicting metric values

jacksontj:
Rebase of #74, fixes #63 

**TLDR; statsd exporter currently doesn't allow for duplicate/conflicting tag-sets -- which is a problem for the statsd exporter. The exporter interface has no such restriction, its simply an issue with the go prometheus client. This can be fixed by implementing our own registry-- which is this PR.**

Apologies for the long delay, other things came up as higher priority. So here is a rebased version against current master to start the conversation from.

As far as the outstanding comments from the previous PR:

`Please update the help string for both flags to make the difference very very clear.` -- Updated with the most recent commit on the branch

`What happens when these are exceeded? What happens for someone with very many metrics?` -- this is a copy/pasted hard-coded value from the [prometheus client_golang](https://github.com/prometheus/client_golang/blob/master/prometheus/registry.go#L31) when the channel hits capacity it will block collection until it has been consumed (effectively a memory limiter).

`that PR is closed. what's the future here?` -- We'll have to maintain this registry long-term. If we'd prefer I can pull the registry into a separate package (directory in the same repo).

cc: @matthiasr

### DIFF
--- a/registry.go
+++ b/registry.go
@@ -1,0 +1,328 @@
+// Copyright 2014 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// TODO: split into separate package
+
+package main
+
+import (
+	"bytes"
+	"fmt"
+	"sort"
+	"sync"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/prometheus/client_golang/prometheus"
+
+	dto "github.com/prometheus/client_model/go"
+)
+
+const (
+	// Capacity for the channel to collect metrics and descriptors.
+	capMetricChan = 1000
+	capDescChan   = 10
+)
+
+var (
+	registry = NewMetricsRegistry()
+)
+
+// For statsd all our metrics will be both a Metric and a collector -- to make life easier
+// we'll define an interface
+type CollectorMetric interface {
+	prometheus.Metric
+	prometheus.Collector
+}
+
+// TODO: remove once https://github.com/prometheus/client_golang/pull/326 is merged
+type Metric struct {
+	// TODO: remove once these have getters upstream
+	Name string
+	Help string
+
+	M CollectorMetric
+}
+
+// TODO: change back to using the regular mechanisms to determine collisions after https://github.com/prometheus/client_golang/pull/326 is merged
+// then we can actually implement the registerer interface
+type MetricsRegistry struct {
+	mtx           sync.RWMutex
+	metricsByID   map[uint64]*Metric // ID is a hash of the descIDs.
+	metricsByDesc map[*prometheus.Desc]*Metric
+}
+
+// NewRegistry creates a new vanilla Registry without any Collectors
+// pre-registered.
+func NewMetricsRegistry() *MetricsRegistry {
+	return &MetricsRegistry{
+		metricsByID:   map[uint64]*Metric{},
+		metricsByDesc: map[*prometheus.Desc]*Metric{},
+	}
+}
+
+// TODO: (deosn't work now) Register implements Registerer.
+func (r *MetricsRegistry) Register(metricID uint64, m *Metric) error {
+	r.mtx.Lock()
+	defer r.mtx.Unlock()
+	if _, exists := r.metricsByID[metricID]; exists {
+		/* TODO?
+		return prometheus.AlreadyRegisteredError{
+			ExistingCollector: existing,
+			NewCollector:      c,
+		}
+		*/
+		return fmt.Errorf("Duplicate metric hash")
+	} else {
+		r.metricsByID[metricID] = m
+		r.metricsByDesc[m.M.Desc()] = m
+		return nil
+	}
+}
+
+// TODO: (deosn't work now) Unregister implements Registerer.
+func (r *MetricsRegistry) Unregister(metricID uint64) bool {
+	r.mtx.RLock()
+	defer r.mtx.RUnlock()
+	if existingMetric, exists := r.metricsByID[metricID]; exists {
+		delete(r.metricsByID, metricID)
+		delete(r.metricsByDesc, existingMetric.M.Desc())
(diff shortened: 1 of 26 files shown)


registry.go
```diff
@@ -0,0 +1,328 @@
...
+
+	// Scatter.
+	// (Collectors could be complex and slow, so we call them all at once.)
+	wg.Add(len(r.metricsByID))
+	go func() {
+		wg.Wait()
+		close(metricChan)
+	}()
+	for _, metric := range r.metricsByID {
+		go func(collector prometheus.Collector) {
```
matthiasr: As pointed out in https://github.com/prometheus/statsd_exporter/issues/76#issuecomment-359783877 this causes a huge spike in memory consumption because it spawns a goroutine for every single time series.

prometheus/client_golang#370 is a potential fix for that.

jacksontj:
If we end up needing to continue with the registry we can either optimize
or implement a more naive one. I'd be more than happy to not "build our
own" but the response last time this was discussed upstream was basically
"try it yourself and of we like it we might merge it".

beorn7:
Re-implementing the registry to create inconsistent metrics sounds like a really bad idea, to be honest.
The ideas for https://github.com/prometheus/client_golang/issues/47 and https://github.com/prometheus/client_golang/issues/355 are around allowing collectors to be declared "unchecked", and then the registry in the client library will try its best to pad missing labels (using empty label values). This will guarantee that every single exposition is consistent, but it will not guarantee consistency over the lifetime of a binary (which is probably a price we can/need to pay to get out of this coundrum).

matthiasr:
So we just discussed this separately – the linked issues don't quite cut it for the statsd exporter. They work well with custom collectors that generate ConstMetrics, but this exporter doesn't. It could somewhat sensibly for counters and gauges, but we'd have to recreate the quantile calculation from scratch for summaries, and the bucketing for histograms.

Another idea that we've kicked around is tracking the labels for a given metric name (this would be easy on top of #119), and when they change unregister the metric and register a completely new one. That's not allowed at the moment anyway, but could be without breaking the semantics too much. On top of that, we could fill out with empty labels as needed anyway. This would mean that at the transition, we would throw away all previous observations (unless we dig into the protobuf to rescue them and copy them over into the new metrics somehow.)

There's no really pretty solution yet …

beorn7:
More thoughts:

The "unregister" hack could be implemented more cleanly (and without changing the semantics of `Unregister`) by using a local (not the global) registry and simply register all metrics (the newly labeled and the old ones that haven't changed) with the new registry. Then throw the old HTTP handler and the old registry away and wire up the new registry.

WRT throwing away the old observations: With summaries, you want a decay time anyway (by default 10m), so assuming label changes are rare, that's not a big deal. For histogram, throwing away the old observations is essentially a counter reset and will be dealt with by Prometheus, i.e. you'll only lose about half a scrape interval worth of even counts.

I'd say let's go with the "throw away on label change" approach, i.e. whenever a previously unaccounted label shows up, recreate all affected metrics with a zero count in a new registry, register all unaffected metrics with the new registry, rewire the HTTP handle to the new registry and throw away the old one.

Does that make sense?

matthiasr:
Could there be any strange side effects from not resetting all metrics at
the same time? Say, for request totals and errors?

brian-brazil:
Technically no, but I can imagine that causing artifacts. On the other hand applications should not be changing labels at runtime.

matthiasr:
The statsd exporter is not, by itself, an application – I would still recommend coupling its lifecycle to that of the application but I also recognize that that is _not_ making it a drop-in replacement for statsd (where a central deployment is much more common). Now, with the schema @beorn7 proposed, we would re-register and reset the metric that changed, but we would not have a mechanism to do that to _all_ metrics exposed by one application at the same time (because we don't know what these all are).

I'm not saying it's a bad idea, just writing down thoughts about the limitations. Whatever we choose to do, I'd want to document the different choices and tradeoffs really well. We could even make the metrics-reregistering an optional feature (and drop the event otherwise).

beorn7:
> Could there be any strange side effects from not resetting all metrics at the same time? Say, for request totals and errors?

To quote from my Prometheus proverb collection: “First do `rate`, then aggregate.” That means, that counter resets are taken care of before aggregating. I guess the scenario you have in mind is a counter for total, that doesn't change its labels, while suddenly error counters arrive with an additional label. Thus, the error counters get a reset but not the total counter. With the usual exception of losing a few increments, it should be fine.

matthiasr:
@jacksontj is this (wiping the whole metric and recreating with the new label set) a solution that would fit your needs?

grobie:
How would that work out during a deployment, where some instances send the new label set and others still the old one? If I understand the proposal correctly, the exporter would constantly throw away the existing metric and create a new one.

matthiasr:
I wouldn't do this for shrinking label sets, those I would just pad out indefinitely. So it would only happen once per metric name, the first time an event with the wider label set is seen. Renaming a label would result in a "wider" label set that includes both.

jacksontj:
@matthiasr that would mean that the metrics sent from clients to this exporter would have labels *added* with empty values. IMO thats not workable, as that means I can no longer query based on the existence of label keys -- since they are being mutated before storage. 

the intention of this exporter is to be (as much as possible) a drop-in-replacment for a statsd server/aggregator/relay/etc. With that in mind we should avoid mutating the metrics as much as possible-- we want to store exactly what we where sent.

matthiasr:
Fundamentally, that's an impossible task. The data models are not the same so we can't store exactly what we get sent – only Graphite can.

The Prometheus server does not distinguish between empty and non-present labels, so you can't query based on label key presence anyway – the query language does not allow that. The only way to test for "label is not present" is to match `{label=""}` anyway.

matthiasr:
I'm going to close this – the concrete changeset is very old, and I believe with prometheus/client_golang#425 we can solve this without a whole new registry.

The concrete problem in #63 was fixed from another angle.

PR #123, opened 2018-02-06
Set counters instead of adding them

### LINKED ISSUE #122: Counter metrics are being added instead of replaced
I have a statsd producer that's emitting a constant `counter` type metric

i.e:

```
myhost.myexec.http.status.404.count:2329|c
myhost.myexec.http.status.200.count:30|c
myhost.myexec.http.status.304.count:10|c
```

Because the metric is a `counter` in the origin system, if I don't get any requests during the scraping period, it'll send again the same value the next time.

I've added the following mapping in my statsd_exporter to use prometheus dimentions like this:

```yaml
mappings:
- match: ^([^.]*)\.([^.]*)--http.status.([^.]*).count
  match_type: "regex"
  name: "http_status_count"
  labels:
    code: "$3"
```

So the ending metric is something like: 
```
http_status_count{code="200"} 1557
http_status_count{code="304"} 18
http_status_count{code="404"} 6987
```

What's happening is that statsd_exporter is constantly *adding* the values that the origin is pushing even though they don't change. So what I end up seing in the prometheus metrics is a constant increment even though the metrics don't change. 

Any ideas?

marcosnils:
Fixes #122

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -288,7 +288,7 @@ func (b *Exporter) Listen(e <-chan Events) {
 					help,
 				)
 				if err == nil {
-					counter.Add(event.Value())
+					counter.Set(event.Value())
 
 					eventStats.WithLabelValues("counter").Inc()
 				} else {
--- a/exporter_test.go
+++ b/exporter_test.go
@@ -115,6 +115,59 @@ func TestHistogramUnits(t *testing.T) {
 	}
 }
 
+type MockCounter struct {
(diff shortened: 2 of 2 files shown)


matthiasr:
As commented in #122, and stated in [the protocol description](https://github.com/etsy/statsd/blob/master/docs/metric_types.md#counting), adding is the correct behavior for statsd counters, and setting would be incorrect.

PR #136, opened 2018-06-15
Allow mapping match on metric type

westphahl:
StatsD allows emitting different metric types under the same metric name, but the Prometheus client library can't merge those. For those use-cases the mapping definition now allows you to specify which metric type to match.

Resolves: #63


matthiasr:
This is wonderful, thank you very much!

PR #140, opened 2018-08-12
Move mapper to a package

matthiasr:
This allows the mapping logic and configuration format to be re-used by
the graphite exporter.

Enables prometheus/graphite_exporter#37, cf. https://github.com/prometheus/graphite_exporter/pull/52#issuecomment-412324849


mapper/telemetry.go
```diff
@@ -0,0 +1,29 @@
...
+
+package mapper
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
+	mappingsCount = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "statsd_exporter_loaded_mappings",
```
matthiasr: i am not sure what to do about this. I could change the metric name to be package- not exporter-specific, or change the constructor to get it injected from the main package.

travisgroth: I'm thinking return the value during load, or keep track of it as a regular value internally.  The caller can then fetch the count and construct the metric however it wants.  This is probably cleaner than creating a new metric in what is effectively a config and parsing package.

In the statsd_exporter case, the prometheus metric would move into main/telemetry and you'd update it whenever the file is loaded.

travisgroth:
Should probably tuck this all into a top level pkg/ directory since the intent is to allow the mapper to be consumed by semi-external projects.

mapper/telemetry.go
matthiasr: Hmm I don't like the scrape-time plumbing that this implies in every user of the package, and that would need to update with every change of telemetry. The package also already knows about Prometheus and imports the client library, so I think we can let main just pass in the Gauge. That's very flexible for users of the package, but also simple if you just want the damn metrics to work.

matthiasr:
@travisgroth I moved everything under pkg/ and we're no longer initializing the gauge from within the library. PTAL!

mapper/telemetry.go
travisgroth: Makes sense to me 👍

PR #142, opened 2018-08-20
Label conflicting events by event name

negz:
This should help diagnose _which_ events are conflicting, at the expense of increasing cardinality. My intuition is that relatively small sets of event names should end up conflicting over time.

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -310,7 +310,7 @@ func (b *Exporter) Listen(e <-chan Events) {
 					eventStats.WithLabelValues("counter").Inc()
 				} else {
 					log.Debugf(regErrF, metricName, err)
-					conflictingEventStats.WithLabelValues("counter").Inc()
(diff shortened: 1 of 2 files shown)


matthiasr:
I would really rather not do this. It is possible to template the metric name, so there is a lot of potential for blowing up not just the exporter but any Prometheus scraping it as well. And once that happens it's very hard to find out what is going on because you're suddenly flying blind.

Instead of adding it to the metric, what do you think about raising the level of the log message so that it is shown by default?

negz:
Fair enough! Logging it at info or warning level is a good compromise, though to be honest debug isn't that bad now that it can be turned on via the CLI. :) On one hand I worry about log spam if it's logging above debug, on the other hand doing so would have helped me determine much faster why my metric was not updating.

matthiasr:
Makes sense, I'll close this for now. I filed #143 for thinking more about the log levels.

PR #152, opened 2018-08-31
Add 'ip' label containing the IP of the reporter

costinm:
Without the IP, reports from multiple sources override each other. 
Probably needs some configuration - but not familiar enough with the codebase, hopefully someone can 
take this over. I'm using this in Istio to identify reports from different envoys.

### DIFF
--- a/bridge_test.go
+++ b/bridge_test.go
@@ -288,6 +288,7 @@ func TestHandlePacket(t *testing.T) {
 			}
 
 			for j, expected := range scenario.out {
+				delete(actual[j].Labels(), "ip")
 				if !reflect.DeepEqual(&expected, &actual[j]) {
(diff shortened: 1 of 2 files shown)


matthiasr:
Before you put more work into this, I'm afraid I have to say no to this feature overall.

In the classic statsd setup with a central receiver, statsd does not do this either.

Some applications send their own identity as part of the metric name (we overloaded Graphite at SoundCloud this way, just before switching to native Prometheus instrumentation). You can use the mapping configuration to put that into a label.

My preferred way of deploying this exporter is 1:1 with application processes, and in that case the Prometheus service discovery already gets you an instance label.

There are multiple ways of getting an identity of the source of the metrics, all with more control than just taking the IP. I do not want to add another one.

PR #157, opened 2018-09-10
Faster glob matching using a finite state machine

fffonion:
This patch proposes a fast and efficient matching algorithm that as a drop in replacement to existing glob matching. On our production environment, the go regexp engine which glob matching uses takes a lot more CPU time than system calls under heavy workloads. We have deployed this patch on our environment and seeing a great performance boost. See [this blog](https://konghq.com/blog/optimizing-prometheus-statsd-exporter-for-scale/) if you are interested with the background : )

By default, a general string takes higher priority than `*` when matching. Backtrack will run under the following circumstances:

- The rules needing backtrack to be mapped correctly. A warning log like `WARN[0000] backtracking required because of match "test2.*.*",matching performance may be degraded
  source="mapper.go:418"` will be printed.
- `glob_disable_ordering` is set to false or unset.

Full changelog:

- Add a new cli argument `statsd.dump-fsm`  to dump the current state machine to a [GNU dot](https://www.graphviz.org/) file. This can help to debug the behaviour of internal state machine.
- Add a new config `glob_disable_ordering` to `defaults` section to indicate that the rules are not strictly in the order of it's occurrence in the mapping config. Setting this to on will boost matching performance under certain large amount of sets of rules.


matthiasr:
Hmm hmm, this is very intriguing. I am not sure about adding another match type that is so similar to glob … is it theoretically possible to make the FSM method fully compatible with the current behavior of the glob type? What would be the performance impact of that be? Where fallbacks are necessary, can this be automatically detected (e.g. whether backtracking is necessary?)

matthiasr:
… what I am thinking is: can we turn this into a full-featured replacement of the current glob implementation? And then, if necessary, add specific settings for further optimizations like treating `*` with lower priority?

matthiasr:
and where ever we can, auto-detect and warn instead of requiring an explicit setting, like

```
WARN: backtracking required for match client.*.request.size.*, falling back to slower glob method
```

(but I would prefer if we could sensibly do the backtracking in the FSM implementation instead of indefinitely carrying around the regex-based-glob implementation)

fffonion:
Yes it's possible to implement the backtrack. Basically it will traverse over all possible transitions until it get one match. The performance impact added will be very little as I can see, since each lookup of transitions is in a map. I like this approach better too, optionally we can detect it when loading as well and also give user an option to do backtracking or not. (And we don't need the `fsm_fallback` anymore)

The ordering of rules can be implemented by replacing the map with array, so that we can keep the order when iterating over it. The performance impact will depend on the types of rules, the more transitions it has for each state, the longer the array will be, and the more the performance impact it will have. Having multiple captures in the label is similar to that and we can basically have a array to store the captureIdx'es and iterate it when formatting. One thing is that once we go this approach, the performance impact on doing backtrack will be magnified.

How does this sound to you:

- By default this `fsm` matcher replaces glob, it uses array to find transitions, having multiple captures and supports backtrack and should pass all the tests for glob matching
- Have a option that satisfy simpler use cases, it uses map to find transitions, can have only one capture per label, and don't support backtrack.

matthiasr:
That sounds perfect! what would be a good name for the latter option?
Should it be one, or a separate toggle for each?

I would like to pull the debug "dump FSM" option out of the mapping
configuration into a flag, just to keep the mapping tidy. That way the file
writing can also be handled by the exporter, instead of mixing it in with
the rather pure mapping logic. Does that sound reasonable?

fffonion:
> That sounds perfect! what would be a good name for the latter option?
Should it be one, or a separate toggle for each?

Hmm it's hard to name it separately but I also feel it's better to separate them from end-user's view. Do you have any suggestions on this?

> I would like to pull the debug "dump FSM" option out of the mapping
configuration into a flag

Yep this sounds good to me.

matthiasr:
Thinking about it, only the ordering is something the mapper needs to know
about ahead of time to preserve performance, right?

if my rules don't require backtracking, the mapper won't spend time on it,
so we don't need a setting for it – maybe log a warning like I sketched out
earlier.

For the label substitution, well, we do the work that the label requires,
so if I only use a simple substitution won't it still be fast? if
absolutely necessary, the mapper could switch modes automatically, but I'd
like to see benchmarks or profiles showing that this complication is
necessary.

For anything that can be handled automatically I'd rather not even add a
configuration knob.

fffonion:
Yep let me try to implement this : ) Thanks for discussion!

fffonion:
Hi @matthiasr, I've finished developing fsm as a drop-in replacement for glob matching. Could you take another look?

Ordering is implemented with minimum performance penalty, still using a map. And we do backtrack only we detect that rules need backtracking. Also I added a test `TestRPS` and shows that the new algorithm is about 33% faster than using regex for glob, when ordering is disabled and no backtrack is needed.

fffonion:
@matthiasr  I've finished the dev part of your comment.

GetMapping func in mapper is cleaned up(no `goto`s 😄) also more comments are added.

FSM parts are moved to a sub package `fsm` inside `mapper` package, because I feel like they are not in a equal level and `mapper` should include `fsm` .

The function dumpFSM is also changed to accept a `io.Writer` instance.

For the performance part, the test func `TestPerformance` is refined and charged with more cases. In short, fsm will have best performance if the rules amount is not so less and captures (use of *) is not to much in a single rule. Ordering and backtracking won't hurt performance much.

I'll start working on the doc. Would you like that to be a separate PR or inside this one with separate commits?

fffonion:
Thanks! I'll update here when I finished the document.

As for ordering, yes performance wise it won't hurt much in normal use case, compared to the gain it get. Turning on ordering will lead rules that don't actually need backtracking to force backtracking, then in some worst cases it essentially let the fsm iterate over the whole tree. Per the performance test, the list assign of captured group is far more expensive than finding a transition. Like the following rules with ordering enabled:

```
a.*.*.**
a.b.*.*.*
a.b.c.*.*
a.b.c.d.*
```
To match `a.b.c.d.e`, the fsm will need to do 10 times of list assign, because it will first hit `a.b.c.d.*`, then `a.b.c.*.*`. While ordering is disabled, it will need at most 4 list assign. This might be a very bad use case, but personally I feel there's still a need to let user choose if they really want that behaviour or not?

matthiasr:
yep, makes sense to keep it then. and please add this example to the docs :)

matthiasr:
I renamed the PR to reflect the direction it has taken

fffonion:
I'll reflect those in the doc : ) Awesome, thanks!

pkg/mapper/fsm/fsm.go
```diff
@@ -0,0 +1,375 @@
...
+			}
+
+			// goto next state
+			root = state
+		}
+		finalStates = append(finalStates, root)
+	}
+	nameFmt := newTemplateFormatter(name, captureCount)
+
+	currentLabelFormatter := make(map[string]*templateFormatter, captureCount)
```
matthiasr: it feels awkward that the FSM logic needs to know about "labels" and "names" … would it make sense to promote the whole formatting business into the mapper package, and make the API of the fsm package such that it can do that? then from the view of the FSM, this would all be subsumed under the "result" field.

pkg/mapper/fsm/fsm.go
```diff
@@ -0,0 +1,375 @@
...
+			}
+		}
+		idx++
+	}
+	// color for start state
+	w.Write([]byte(fmt.Sprintf("0 [color=\"#a94442\",fillcolor=\"#f2dede\"];\n")))
+	w.Write([]byte("}"))
+}
+
+// TestIfNeedBacktracking test if backtrack is needed for current FSM
```
matthiasr: the comment suggests that this method only looks at the FSM, but it also takes a slice of strings, what does that parameter mean?

pkg/mapper/mapper.go
```diff
@@ -160,6 +184,19 @@ func (m *MetricMapper) InitFromYAMLString(fileContents string) error {
...
 	m.Defaults = n.Defaults
 	m.Mappings = n.Mappings
+	if n.doFSM {
+		var mappings []string
+		for _, mapping := range n.Mappings {
+			if mapping.MatchType == MatchTypeGlob {
+				mappings = append(mappings, mapping.Match)
+			}
+		}
+		n.FSM.TestIfNeedBacktracking(mappings)
```
matthiasr: what happens to the value of this? as it is, isn't this a NOOP?

pkg/mapper/fsm/fsm.go
```diff
@@ -0,0 +1,375 @@
...
+						break
+					}
+				}
+			}
+
+			for i2, r2 := range rules {
+				if i2 != i1 && len(re1.FindStringSubmatchIndex(r2)) > 0 {
+					// log if we care about ordering and the superset occurs before
+					if !f.disableOrdering && i1 < i2 {
+						log.Warnf("match \"%s\" is a super set of match \"%s\" but in a lower order, "+
```
matthiasr: for clarity: `match \"%s\" is a super set of match \"%s\"  but comes first, it will never be matched`

matthiasr: I don't think you need `\n` when logging

pkg/mapper/fsm/fsm.go
```diff
@@ -0,0 +1,375 @@
...
+					"matching performance may be degraded\n", r1)
+				needBacktrack = true
+			}
+		}
+	}
+
+	// backtracking will always be needed if ordering of rules is not disabled
+	// since transistions are stored in (unordered) map
+	// note: don't move this branch to the beginning of this function
+	// since we need logs for superset rules
```
matthiasr: we only log if `!f.disableOrdering`, and we only honor the result of the computation otherwise. would it make sense to split this into two methods:

* one that, assuming ordered matches, logs warnings for dead rules
* one that, assuming unordered matches, calculates whether backtracking is needed

then call one or the other, depending on the disableOrdering configuration?

pkg/mapper/fsm/fsm.go
```diff
@@ -0,0 +1,375 @@
...
+	f.needsBacktracking = !f.disableOrdering || needBacktrack
```
matthiasr: please _either_ change f (imperative style) _or_ return something (functional style) and set in the caller. I prefer the latter.

matthiasr: as far as I can see, if you split "log warnings for ordered mappings" from "calculate backtracking need for unordered mappings", then neither method needs the FSM itself?

pkg/mapper/fsm/fsm.go
```diff
@@ -0,0 +1,375 @@
...
+		state.resultNameFormatter = nameFmt
+		state.resultLabelsFormatter = currentLabelFormatter
+		state.resultPriority = f.statesCount
+	}
+
+	f.statesCount++
+
+}
+
+// DumpFSM accepts a io.writer and write the current FSM into dot file format
```
matthiasr: maybe move this method to its own file? it's fairly extra from everything else

pkg/mapper/mapper.go
matthiasr: I see now, the method also mutates the FSM, I would prefer not to do that – see my comment there.

pkg/mapper/mapper.go
```diff
@@ -75,6 +81,20 @@ var defaultQuantiles = []metricObjective{
 	{Quantile: 0.99, Error: 0.001},
 }
 
+func min(x, y int) int {
```
matthiasr: I think these are moved into the FSM now too?

pkg/mapper/mapper_test.go
```diff
@@ -518,6 +608,538 @@ mappings:
...
 
+func duplicateRules(count int, template string) string {
+	rules := ""
+	for i := 0; i < count; i++ {
+		rules += fmt.Sprintf(template, i)
+	}
+	return rules
+}
+
+func TestPerformance(t *testing.T) {
```
matthiasr: Please write this in terms of a [Go benchmark](https://golang.org/pkg/testing/#hdr-Benchmarks) – it handles the number of iterations automatically for us. Also move it to a separate file, just because this one is so terribly long already.

matthiasr:
Thank you for continuing to work on this! I am sorry for all the comments and back and forth – this is a pretty massive change and I am a bit scared that we won't be able to understand and improve it later unless we get it really well structured and documented now.

fffonion:
No problem, I totally understand that.: ) I'll address the comments and finish the docs, it will take some time but I'll let you know when I get there.

main.go
```diff
@@ -118,6 +120,20 @@ func watchConfig(fileName string, mapper *mapper.MetricMapper) {
 	}
 }
 
+func dumpFSM(mapper *mapper.MetricMapper, dumpFilename string) error {
+	f, err := os.Create(dumpFilename)
```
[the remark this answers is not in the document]
fffonion: It will rewrite the existing file, I feel this is acceptable?

main.go
```diff
@@ -126,6 +142,7 @@ func main() {
 		statsdListenTCP = kingpin.Flag("statsd.listen-tcp", "The TCP address on which to receive statsd metric lines. \"\" disables it.").Default(":9125").String()
 		mappingConfig   = kingpin.Flag("statsd.mapping-config", "Metric mapping configuration file name.").String()
 		readBuffer      = kingpin.Flag("statsd.read-buffer", "Size (in bytes) of the operating system's transmit read buffer associated with the UDP connection. Please make sure the kernel parameters net.core.rmem_max is set to a value greater than the value specified.").Int()
+		dumpFSMPath     = kingpin.Flag("statsd.dump-fsm", "The path to dump internal FSM generated for glob matching as Dot file.").Default("").String()
```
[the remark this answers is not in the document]
fffonion: Yeah i second on the debug.dump-fsm, let me change it.

pkg/mapper/mapper.go
```diff
@@ -31,18 +31,24 @@ var (
 	metricLineRE = regexp.MustCompile(`^(\*\.|` + statsdMetricRE + `\.)+(\*|` + statsdMetricRE + `)$`)
 	metricNameRE = regexp.MustCompile(`^([a-zA-Z_]|` + templateReplaceRE + `)([a-zA-Z0-9_]|` + templateReplaceRE + `)*$`)
 	labelNameRE  = regexp.MustCompile(`^[a-zA-Z_][a-zA-Z0-9_]+$`)
+
+	templateReplaceCaptureRE = regexp.MustCompile(`\$\{?([a-zA-Z0-9_\$]+)\}?`)
```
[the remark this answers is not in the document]
fffonion: no it's not needed, will cleanup

pkg/mapper/fsm/formatter.go
```diff
@@ -0,0 +1,68 @@
...
+		// if no regex reference found, keep it as it is
+		return &templateFormatter{captureCount: 0, fmtString: valueExpr}
+	}
+
+	var indexes []int
+	valueFormatter := valueExpr
+	for _, match := range matches {
+		idx, err := strconv.Atoi(match[len(match)-1])
+		if err != nil || idx > captureCount || idx < 1 {
+			// if index larger than captured count or using unsupported named capture group,
```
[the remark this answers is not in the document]
fffonion: it's only supported in regex, although glob used regex under the hood it was not possible to name the group

main.go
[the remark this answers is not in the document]
matthiasr: yes, sounds good!

pkg/mapper/fsm/fsm.go
```diff
@@ -0,0 +1,375 @@
...
+	metricTypes       []string
+	disableOrdering   bool
+	statesCount       int
+}
+
+// NewFSM creates a new FSM instance
+func NewFSM(metricTypes []string, maxPossibleTransitions int, disableOrdering bool) *FSM {
+	fsm := FSM{}
+	root := &mappingState{}
+	root.transitions = make(map[string]*mappingState, 3)
```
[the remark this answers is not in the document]
fffonion: sorry magic numbers ;), it's three metrics types we currently supporting so the initial transitions are three. I'll change to use len() instead.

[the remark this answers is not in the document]
fffonion: sorry magic numbers ;), it's three metrics types we currently supporting so the initial transitions are three. I'll change to use len() instead.

pkg/mapper/fsm/fsm.go
```diff
@@ -0,0 +1,375 @@
...
+	matchFields := strings.Split(match, ".")
+	// fill into our FSM
+	roots := []*mappingState{}
+	if matchMetricType == "" {
+		// if metricType not specified, connect the state from all three types
+		for _, metricType := range f.metricTypes {
+			roots = append(roots, f.root.transitions[string(metricType)])
+		}
+	} else {
+		roots = append(roots, f.root.transitions[matchMetricType])
```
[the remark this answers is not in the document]
fffonion: i guess the yaml unmarshaller will raise an error in that case (https://github.com/prometheus/statsd_exporter/blob/master/pkg/mapper/metric_type.go#L26)

pkg/mapper/fsm/fsm.go
fffonion: yeah it makes more sense, and we don't need to pass those variables around. let me try to implement that.

pkg/mapper/fsm/fsm.go
fffonion: emm this is not an accurate comment, let me tune on it

pkg/mapper/mapper.go
fffonion: oh yeah, sorry 😅

pkg/mapper/fsm/fsm.go
fffonion: yes, i'll fix this along with the other comment

pkg/mapper/mapper.go
```diff
@@ -94,7 +114,14 @@ func (m *MetricMapper) InitFromYAMLString(fileContents string) error {
...
 
+	remainingMappingsCount := len(n.Mappings)
+
+	n.FSM = fsm.NewFSM([]string{string(MetricTypeCounter), string(MetricTypeGauge), string(MetricTypeTimer)},
+		remainingMappingsCount, n.Defaults.GlobDisableOrdering)
+
 	for i := range n.Mappings {
+		remainingMappingsCount--
+
 		currentMapping := &n.Mappings[i]
```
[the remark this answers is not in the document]
fffonion: I think there's some mutation to currentMapping that will be broken if we use take currentMapping from `range`, which is not a pointer.

matthiasr:
almost forgot: please also add `go test -bench .` to the TravisCI build so we can continuously see the results there. thank you!

fffonion:
Hi @matthiasr, I've finished the doc and addressed your comments! Could you take another look please?

fffonion:
Thank you @matthiasr for reviewing and taking care of this!

PR #162, opened 2018-11-02
Break out event handling into its own function

grobie:
It's idiomatic in go to keep code indentation due to nested loops to a
minimum. Decoupling the events handling function into its own function
makes it easier to read the code and test the behavior in isolation of
the channel handling. It's now also easily possible to process events
in parallel without having to touch actual business code.

Signed-off-by: Tobias Schmidt <tobidt@gmail.com>

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -253,6 +253,8 @@ func escapeMetricName(metricName string) string {
 	return metricName
 }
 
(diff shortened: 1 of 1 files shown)


grobie:
https://github.com/prometheus/statsd_exporter/pull/162/files?w=1

Issue #129, opened 2018-03-21
How to identify stale data?


InKolev:
Hi, I see that after I **stopped sending** a given metric **"X"** to statsd for **more than 30 minutes now**,  the **metric still appears in the export (/metrics) page** with its last set value.

I want the statsd_exporter to **stop reporting this metric** after it was stale for a given time.

What is the expected behavior in such cases? 
When will the metric stop being reported?
Is there a way to configure such timers?

Thanks in advance.

matthiasr:
At the moment, there is no way to expire it except by restarting the exporter. I am hesitant to add the complexity of handling expiration; The Prometheus Way™ is to tie the lifecycle of metrics to that of the application – for the statsd exporter that means an instance of the exporter alongside each application instance, restarted together.

xkisu:
I experienced the same thing and originally thought it was a bug, not the intended operation. Restarting the exporter in a container environment is simple enough, but what about when it's running on a VPS with other services on the same VPS feeding it metrics that need to restart from time to time?

It would be nice if there was at least an optional flag to set a timeout to mark a metric as stale.

matthiasr:
Should this be global or per-metric?

matthiasr:
(or rather, per-mapping)

xkisu:
I was thinking global as that's how it would best apply to most of the use cases I'm thinking about, but I can see how per-metric could be useful too

matthiasr:
Makes sense. We can probably start global and see if that solves the issues. So if anyone wants to implement this, please comment here and submit a PR!

brian-brazil:
I'm not sure that's going to solve anything - if anything it'll just cause more problems. A metric is still valid even if it hasn't been updated in a long time - e.g. process_start_time_seconds.

brian-brazil:
It's not a problem if the standard architecture is followed with the exporter matching the lifecycle of the thing that's exporting to it. This suggestion violates https://prometheus.io/docs/practices/instrumentation/#avoid-missing-metrics

matthiasr:
In a *Prometheus* "standard architecture", you don't need this exporter.

I see the purpose of this as providing a bridge between applications instrumented for a *statsd/graphite* standard architecture to metrics collection in Prometheus. These applications do not, semantically, follow the Prometheus instrumentation model, so if they emit a process start time, they will do so continuously, because that's what you need to do to get a continuous line in Graphite. Whether it's a gauge or not has no impact on this at all.

[...]

PR #164, opened 2018-11-13
Remove stale metrics

diafour:
This patch implements clearing of metrics with variable labels with per-mapping configurable timeout (ttl). Our monitoring installation is suffered from the behavior desribed in #129. There are metrics from Kubernetes pods with constantly changing values for labels. For example, metric `ingress_nginx_upstream_retries_count` with label `pod` and changing pod name as a value for this label. There is no sense to store metrics with old pod names as they are never repeated. Removing metrics with this stale values will not cause any damage to subsequent aggregations or multiple prometheus servers.

The patch is divided by 3 parts to simplify a review:
- replace Metrics with Collectors to gain ability of deletion metrics. So Elements in CounterContainer stores CounterVecs instead of Counters.
- implement labels values storage in Exporter to detect metrics staleness
- implement configurable timeout (ttl). Add key `ttl` to `default` section and to `mapping` sections.

`ttl` is a timeout in seconds for stale metrics. Exporter saves each labels values set and updates last time it receives metric with these labels values. If no metric with the same labels values received for `ttl` seconds, then statsd_exporter stops reporting metric with this labels values set.

Mark this as WIP. It would be great to hear your thoughts on this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,6 +268,7 @@ defaults:
   buckets: [.005, .01, .025, .05, .1, .25, .5, 1, 2.5 ]
   match_type: glob
   glob_disable_ordering: false
+  ttl: 0 # metrics not expired
 mappings:
 # This will be a histogram using the buckets set in `defaults`.
 - match: test.timing.*.*.*
@@ -349,6 +350,18 @@ mappings:
 ```
 
 Possible values for `match_metric_type` are `gauge`, `counter` and `timer`.
+
+### Time series expiration
+
+`ttl` parameter can be used to define expiration time for stale metrics.
+Value is a time duration with valid time units: "ns", "us" (or "µs"),
+"ms", "s", "m", "h". For example, `ttl: 1m20s`. `0` value is used to indicate
+not expired metrics.
+
+Expiration is useful to gather metrics with changing label values from
+dynamic environments like Kubernetes. For example, metric
+`ingress_nginx_upstream_retries_count` with label `pod` and
+changing pod name as a value for this label.
 
 ## Using Docker
 
--- a/README.md
+++ b/README.md
@@ -268,7 +268,7 @@ defaults:
   buckets: [.005, .01, .025, .05, .1, .25, .5, 1, 2.5 ]
   match_type: glob
   glob_disable_ordering: false
-  ttl: 0 # metrics not expired
+  ttl: 0 # metrics do not expire
 mappings:
 # This will be a histogram using the buckets set in `defaults`.
 - match: test.timing.*.*.*
@@ -353,15 +353,15 @@ Possible values for `match_metric_type` are `gauge`, `counter` and `timer`.
 
 ### Time series expiration
 
-`ttl` parameter can be used to define expiration time for stale metrics.
-Value is a time duration with valid time units: "ns", "us" (or "µs"),
+The `ttl` parameter can be used to define the expiration time for stale metrics.
+The value is a time duration with valid time units: "ns", "us" (or "µs"),
 "ms", "s", "m", "h". For example, `ttl: 1m20s`. `0` value is used to indicate
-not expired metrics.
+metrics that do not expire.
 
-Expiration is useful to gather metrics with changing label values from
-dynamic environments like Kubernetes. For example, metric
-`ingress_nginx_upstream_retries_count` with label `pod` and
-changing pod name as a value for this label.
+ TTLs are applied to each mapped metric name/labels combination whenever
+ new samples are received. This means that you cannot immediately expire a
+ metric only by changing the mapping configuration. At least one sample must
+ be received for updated mappings to take effect.
 
 ## Using Docker
 
--- a/exporter.go
+++ b/exporter.go
@@ -50,7 +50,7 @@ var (
 	intBuf = make([]byte, 8)
 )
 
-func labelsNames(labels prometheus.Labels) []string {
+func labelNames(labels prometheus.Labels) []string {
 	names := make([]string, 0, len(labels))
 	for labelName := range labels {
 		names = append(names, labelName)
@@ -91,7 +91,7 @@ func (c *CounterContainer) Get(metricName string, labels prometheus.Labels, help
 		counterVec = prometheus.NewCounterVec(prometheus.CounterOpts{
 			Name: metricName,
 			Help: help,
-		}, labelsNames(labels))
+		}, labelNames(labels))
 		if err := prometheus.Register(counterVec); err != nil {
 			return nil, err
 		}
@@ -122,7 +122,7 @@ func (c *GaugeContainer) Get(metricName string, labels prometheus.Labels, help s
 		gaugeVec = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 			Name: metricName,
 			Help: help,
-		}, labelsNames(labels))
+		}, labelNames(labels))
 		if err := prometheus.Register(gaugeVec); err != nil {
 			return nil, err
 		}
@@ -165,7 +165,7 @@ func (c *SummaryContainer) Get(metricName string, labels prometheus.Labels, help
 				Name:       metricName,
 				Help:       help,
 				Objectives: objectives,
-			}, labelsNames(labels))
+			}, labelNames(labels))
 		if err := prometheus.Register(summaryVec); err != nil {
 			return nil, err
 		}
@@ -204,7 +204,7 @@ func (c *HistogramContainer) Get(metricName string, labels prometheus.Labels, he
 				Name:    metricName,
 				Help:    help,
 				Buckets: buckets,
-			}, labelsNames(labels))
+			}, labelNames(labels))
 		if err := prometheus.Register(histogramVec); err != nil {
 			return nil, err
 		}
--- a/exporter.go
+++ b/exporter.go
@@ -15,17 +15,22 @@ package main
 
 import (
 	"bufio"
+	"bytes"
+	"encoding/binary"
 	"fmt"
+	"hash/fnv"
 	"io"
 	"net"
 	"regexp"
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 	"unicode/utf8"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/log"
+	"github.com/prometheus/common/model"
 
 	"github.com/prometheus/statsd_exporter/pkg/mapper"
 )
@@ -37,8 +42,15 @@ const (
 		"consider the effects on your monitoring setup. Error: %s"
 )
 
+// TODO move to mapping config
+const metricTtl = time.Duration(5 * time.Second)
+
 var (
 	illegalCharsRE = regexp.MustCompile(`[^a-zA-Z0-9_]`)
+
+	hash   = fnv.New64a()
+	strBuf bytes.Buffer // Used for hashing.
+	intBuf = make([]byte, 8)
 )
 
 func labelsNames(labels prometheus.Labels) []string {
@@ -48,6 +60,21 @@ func labelsNames(labels prometheus.Labels) []string {
 	}
 	sort.Strings(names)
 	return names
+}
+
+// hashNameAndLabels returns a hash value of the provided name string and all
+// the label names and values in the provided labels map.
+//
+// Not safe for concurrent use! (Uses a shared buffer and hasher to save on
+// allocations.)
+func hashNameAndLabels(name string, labels prometheus.Labels) uint64 {
+	hash.Reset()
+	strBuf.Reset()
+	strBuf.WriteString(name)
+	hash.Write(strBuf.Bytes())
+	binary.BigEndian.PutUint64(intBuf, model.LabelsToSignature(labels))
+	hash.Write(intBuf)
+	return hash.Sum64()
 }
 
 type CounterContainer struct {
@@ -76,6 +103,12 @@ func (c *CounterContainer) Get(metricName string, labels prometheus.Labels, help
 	return counterVec.GetMetricWith(labels)
 }
 
+func (c *CounterContainer) Delete(metricName string, labels prometheus.Labels) {
+	if _, ok := c.Elements[metricName]; ok {
+		c.Elements[metricName].Delete(labels)
+	}
+}
+
 type GaugeContainer struct {
 	Elements map[string]*prometheus.GaugeVec
 }
@@ -99,6 +132,12 @@ func (c *GaugeContainer) Get(metricName string, labels prometheus.Labels, help s
 		c.Elements[metricName] = gaugeVec
 	}
 	return gaugeVec.GetMetricWith(labels)
+}
+
+func (c *GaugeContainer) Delete(metricName string, labels prometheus.Labels) {
+	if _, ok := c.Elements[metricName]; ok {
+		c.Elements[metricName].Delete(labels)
+	}
 }
 
 type SummaryContainer struct {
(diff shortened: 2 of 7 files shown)


diafour:
Tests are broken for now because of changes in Elements fields. Will fix it at the daylight.

matthiasr:
Awesome, thank you for contributing! It will take me a little while to go through the code changes in detail, but I wanted to say that I think the general approach is good.

exporter.go
```diff
@@ -63,75 +66,133 @@ func hashNameAndLabels(name string, labels prometheus.Labels) uint64 {
...
 	return &GaugeContainer{
-		Elements: make(map[uint64]prometheus.Gauge),
+		Elements: make(map[string]map[uint64]*prometheus.GaugeVec),
 	}
 }
 
 func (c *GaugeContainer) Get(metricName string, labels prometheus.Labels, help string) (prometheus.Gauge, error) {
-	hash := hashNameAndLabels(metricName, labels)
-	gauge, ok := c.Elements[hash]
+	hash := hashNameAndLabelNames(metricName, labels)
```
matthiasr: how is this guarded against concurrent usage?

exporter.go
```diff
@@ -63,75 +66,133 @@ func hashNameAndLabels(name string, labels prometheus.Labels) uint64 {
...
+	}
+	sort.Strings(names)
+	return names
+}
+
+// hashNameAndLabels returns a hash value of the provided name string and all
+// the label names in the provided labels map.
+//
+// Not safe for concurrent use! (Uses a shared buffer and hasher to save on
+// allocations.)
```
matthiasr: could they be taken out of a pool safely? would that be beneficial? what is the consequence of serialised access for ingestion and scrape performance?

exporter.go
```diff
@@ -63,75 +66,133 @@ func hashNameAndLabels(name string, labels prometheus.Labels) uint64 {
...
+func labelsNames(labels prometheus.Labels) []string {
+	names := make([]string, 0, len(labels))
+	for labelName := range labels {
+		names = append(names, labelName)
+	}
+	sort.Strings(names)
+	return names
+}
+
+// hashNameAndLabels returns a hash value of the provided name string and all
```
matthiasr: the name here does not match the function definition

matthiasr: I see that you left `hashNameAndLabels` in – is it still used? why?

exporter.go
```diff
@@ -63,75 +66,133 @@ func hashNameAndLabels(name string, labels prometheus.Labels) uint64 {
...
 			return nil, err
 		}
-		c.Elements[hash] = gauge
+		c.Elements[metricName] = map[uint64]*prometheus.GaugeVec{
+			hash: gauge,
+		}
+	}
+	gauge, ok = c.Elements[metricName][hash]
+	if !ok {
+		err := fmt.Errorf(regErrNames, labelsNames(labels))
```
matthiasr: should we handle the mismatch gracefully at this point? a label set can be "expanded" with empty values without changing semantics.

R0quef0rt:
This is a great idea. Really hoping to see it implemented soon. 

I work in an environment where our central infrastructure is in the cloud, but our clients are behind corporate firewalls. We're developing an application-specific plugin that will send metrics to Prometheus - but we keep running into this problem. 

The pushgateway - and now, statsd - currently "cache" metrics indefinitely. This makes it difficult to determine if a client has gone down - because their metrics are being persisted indefinitely. 

Here's hoping we'll see a TTL feature added soon!

matthiasr:
One workaround for this specific question and situation is to push a metric
that is the current (unix) time from all clients, then you can see when
that stops updating / gets too far away from the server time().

exporter.go
```diff
@@ -63,75 +66,133 @@ func hashNameAndLabels(name string, labels prometheus.Labels) uint64 {
 	return hash.Sum64()
 }
 
+func labelsNames(labels prometheus.Labels) []string {
```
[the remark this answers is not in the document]
diafour: That's right!

exporter.go
diafour: Same as before: methods of Exporter and *Container structs are called synchronously. Each received statsd line is transformed into a set of events. Events are send to a channel in a single instance of Exporter and handled there one by one. 

https://github.com/prometheus/statsd_exporter/blob/master/main.go#L211-L212

https://github.com/flant/statsd_exporter/blob/remove_stale_metrics/exporter.go#L359

exporter.go
diafour: I don't fully get your question but has feeling that partly answered there: https://github.com/prometheus/statsd_exporter/pull/164#discussion_r235991822

exporter.go
diafour: Will clean it up!

exporter.go
diafour: It seems that the only purpose of hashNameAndLabelNames and the second dimension in Elements field is to catch this type of error =)

I got this, need some time to remove unnecessary hashing.

Background story: It would be great if golang prometheus client has lastUpdateTime field for each label values set in MetricVec and has methods to list and delete individual sets. That will reduce memory and cpu footprint. For now there is «write only» behavior and we force to create intermediate layer to store label values.
I've implemented this algorithm in our other product and there is ConstLabels option in use, that is where hash for label names comes from.

matthiasr: Keeping the client library write-only was a deliberate design decision, so we'll have to deal with it either way. The accepted way to solve this without the memory duplication is to implement the Collector interface and send ephemeral ConstMetric objects instead of setting the value on persistent metrics. The graphite exporter [does this](https://github.com/prometheus/graphite_exporter/blob/75f0db17d0be85c83d16855a82acffe31e605179/main.go#L196-L200). I don't expect you to do that refactoring in this PR though, that can happen separately later. Your work actually moves us closer to it.

exporter.go
matthiasr: Ah cool, that makes sense.

exporter.go
diafour: `hashNameAndLabels` is used to calculate a key for label values in Exporter.labelValues.

exporter.go
diafour: OK

diafour:
Tests are good now. Problem was in TestHistogramUnits with mocking of Histogram metric. Resolve it with Write a dto.Metric as in a client_golang https://github.com/prometheus/client_golang/blob/master/prometheus/examples_test.go#L497-L498

`hashNameAndLabels` is used to create a key for label values in intermediate storage (Exporter.labelValues)

`labelNames` is got proper name

diafour:
@matthiasr Any comments? statsd_exporter with this patch now lives on 5 production clusters and works well.

matthiasr:
sorry, I got distracted and then forgot. the title still says WIP?

matthiasr:
What happens when there are two mappings with different TTLs mapping to the same metric name / label? That is a perfectly valid situation, and we need to at least document it.

exporter.go
```diff
@@ -234,12 +262,19 @@ func (c *TimerEvent) MetricType() mapper.MetricType { return mapper.MetricTypeTi
 
 type Events []Event
 
+type LabelValues struct {
+	lastRegisteredAt time.Time
+	labels           prometheus.Labels
+	ttl              uint64
```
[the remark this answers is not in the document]
diafour: I think yes, this can be a `time.Duration`.

exporter_test.go
```diff
@@ -106,14 +97,23 @@ func TestHistogramUnits(t *testing.T) {
...
 	ex.Listen(events)
-	if mock.value == 300 {
+
+	histogram, err := ex.Histograms.Get(name, prometheus.Labels{}, "", nil)
+	if err != nil {
+		t.Fatalf("Histogram not registered")
+	}
+
+	// check the state of the histogram by
+	// (ab)using its Write method (which is usually only used by Prometheus internally).
```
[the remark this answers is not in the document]
diafour: `testutil` is a great missing part!
Internally it uses the same technique with .Write() method and what is bad — it doesn't support Histograms and Summaries for now :(

https://github.com/prometheus/client_golang/blob/master/prometheus/testutil/testutil.go#L97-L107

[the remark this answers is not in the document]
matthiasr: ok fair, in that case leave it as is!

diafour:
> Could you please add tests for the added mapping fields, and tests that demonstrate/verify the expiry behaviour?

Done

> Please update the documentation (README) as well.

Done

> The exporter allows live-reloading of the configuration, so the label set or TTL of a given metric can change during the lifetime of the exporter. What happens in that case?

Modified ttl for a metric will not be in use until call to handleEvent with a new value for metric. In theory one would want to expire a stale metric with `ttl: 30m` without reload by decreasing ttl to 1s — this is not possible now. Should we handle such situation? I think it requires some kind of signalization about config reloads from main to exporter.

> What happens when there are two mappings with different TTLs mapping to the same metric name / label? That is a perfectly valid situation, and we need to at least document it.

`handleEvent` use mapper.GetMapping which use FSM.GetMapping to get a *MetricMapping by metric name and metric type. This mapping defines ttl to use. As I can see labels are not used to find a 
mapping. Can you tell what is an algorithm to find the mapping in FSM? May be you have an example of config to run some tests?

diafour:
@matthiasr some comments about tests. Previous algorithm with getting a Metric instance and call a Write works good for simple cases such as test for 1 value or test values in one routine. So I use a Gatherer which is thread safe and to be more close to what a scraper get. Method `Gather` returns many additional metrics but for test purposes it is not important.

matthiasr:
> Should we handle such situation? I think it requires some kind of signalization about config reloads from main to exporter.

I think for now it's okay to document this and leave it be. If someone really needs instant expiration, they can restart or implement that signalling.

matthiasr:
> Can you tell what is an algorithm to find the mapping in FSM

The FSM finds exactly one mapping for a given statsd metric name. The Prometheus metric name and labels are an output of that. I think what will happen in this case is that whichever mapping last matched will win. Again, I think that's okay as long as we document it.

README.md
```diff
@@ -350,6 +351,18 @@ mappings:
 
 Possible values for `match_metric_type` are `gauge`, `counter` and `timer`.
 
+### Time series expiration
+
+`ttl` parameter can be used to define expiration time for stale metrics.
```
matthiasr:
```suggestion
The `ttl` parameter can be used to define the expiration time for stale metrics.
```

README.md
```diff
@@ -268,6 +268,7 @@ defaults:
   buckets: [.005, .01, .025, .05, .1, .25, .5, 1, 2.5 ]
   match_type: glob
   glob_disable_ordering: false
+  ttl: 0 # metrics not expired
```
matthiasr:
```suggestion
   ttl: 0 # metrics do not expire
```

README.md
```diff
@@ -350,6 +351,18 @@ mappings:
 
 Possible values for `match_metric_type` are `gauge`, `counter` and `timer`.
 
+### Time series expiration
+
+`ttl` parameter can be used to define expiration time for stale metrics.
+Value is a time duration with valid time units: "ns", "us" (or "µs"),
```
matthiasr:
```suggestion
The value is a time duration with valid time units: "ns", "us" (or "µs"),
```

README.md
```diff
@@ -350,6 +351,18 @@ mappings:
...
+"ms", "s", "m", "h". For example, `ttl: 1m20s`. `0` value is used to indicate
+not expired metrics.
```
matthiasr:
```suggestion
metrics that do not expire.
```

README.md
```diff
@@ -350,6 +351,18 @@ mappings:
...
+
+Expiration is useful to gather metrics with changing label values from
```
matthiasr: I would rather not cite Kubernetes (pod) as an example here – in that case I would always recommend running the exporter as a side-car in every pod, so that the pod label can be added by Prometheus at scrape time. That also solves a whole bunch of other label consistency and expiry issues, by tying the life cycles of the application and the exporter together.

Can you delete this paragraph please?

exporter.go
```diff
@@ -390,13 +437,54 @@ func (b *Exporter) handleEvent(event Event) {
...
+	if !ok {
+		b.labelValues[metricName][hash] = &LabelValues{
+			labels: labels,
+			ttl:    ttl,
+		}
+	}
+	now := time.Now()
+	b.labelValues[metricName][hash].lastRegisteredAt = now
+	// Update ttl from mapping
+	b.labelValues[metricName][hash].ttl = ttl
```
matthiasr: you keep separately checking the presence of something in a map (`b.labelValues[metricName]`, `b.labelValues[metricName][hash]`) but discarding the thing itself, then walking into the map again and again.

Would it be better here to keep the values around, and use them directly? Like

```
	metric, hasMetric := b.labelValues[metricName]
 	if !hasMetric {
                metric = make(map[uint64]*LabelValues)
 		b.labelValues[metricName] =metric
 	}
 	hash := hashNameAndLabels(metricName, labels)
 	metricLabelValues, hasMetricLabelValues := b.labelValues[metricName][hash]
        …
```

That way you can save a few map lookups, and I find it easier to follow which value is being used where.

exporter_test.go
```diff
@@ -173,3 +168,211 @@ func TestEscapeMetricName(t *testing.T) {
...
+		stop := make(chan bool, 1)
+
+		l.handlePacket([]byte("foobar:200|g"), events)
+		l.handlePacket([]byte("bazqux.main:42|ms"), events)
+
+		// Close channel to signify we are done with the listener after a short period.
+		go func() {
+			defer close(events)
+
+			time.Sleep(time.Millisecond * 100)
```
matthiasr: Could we test this without _actually_ sleeping (and depending on actual wall clocks and such)? I found an [article](https://smartystreets.com/blog/2015/09/go-testing-part-5-testing-with-time) that makes some good suggestions.

README.md
```diff
@@ -350,6 +351,18 @@ mappings:
...
+dynamic environments like Kubernetes. For example, metric
+`ingress_nginx_upstream_retries_count` with label `pod` and
+changing pod name as a value for this label.
+
```
matthiasr:
```suggestion
 
 TTLs are applied to each mapped metric name/labels combination whenever
 new samples are received. This means that you cannot immediately expire a
 metric only by changing the mapping configuration. At least one sample must
 be received for updated mappings to take effect.
```

matthiasr:
Nevermind my last comment, I was confusing myself about what level TTLs are applied at. I don't expect anyone to have multiple mappings to the same Prometheus metric name, labels _and_ label values, so we don't need to cover that case. If you're really mapping multiple event streams to the same metric, all bets are off, but last-event-wins is what you get either way.

diafour:
I rebase the branch. New prometheus client brings an error "Histogram is not Observer". `Get` methods for Histogram and Summary now return prometheus.Observer to fix this issue.

All changes in README are applied, paragraph about k8s is deleted.

Tests are not depended on time.Sleep:
- time.Now and time.NewTracker are wrapped within clock package to be overridden in tests
- time.Sleep was used to wait for `handleEvent` to finish. Making `events` channel synchronous works well too.

README.md
diafour: Applied

README.md
diafour: Applied

README.md
diafour: Paragraph is deleted

README.md
diafour: Applied

README.md
diafour: Applied

README.md
diafour: Applied

exporter_test.go
diafour: Done

matthiasr:
Thank you very much!

PR #169, opened 2018-12-05
fix: don't let captured fields being overwritten by backtracking

### LINKED ISSUE #168: Backtracking causes wrong mapping of labels
When I have the following mappings defined.
```
---
mappings:
- match: '*.dummy.*'
  name: metric_one
  labels:
    system: $1
    attribute: $2
- match: 'full.name.*'
  name: metric_two
  labels:
    system: static
    attribute: $1
```
And then send in the following 2 metrics
```
whatever.dummy.test
full.name.anothertest
```
I expect the following result
```
metric_one{attribute="test",system="whatever"}
metric_two{attribute="anothertest",system="static"} 
```
But instead I get
```
metric_one{attribute="test",system="whatever"}
metric_two{attribute="full",system="static"} 
```
It's as if after backtracking, it remembers that `$1` mached the first part, `full` - but when going down the second path, the first part is not a `*`, and should not match it - but rather match on the 3rd part `anothertest` - which is now dropped entirely from the result.

I was able to verify this in the `mapper_test.go` by adding the following scenario.
```
		//Config with backtracking, starting star
		{
			config: `
defaults:
  glob_disable_ordering: false
mappings:
- match: '*.dummy.*'
  name: metric_one
  labels:
    system: $1
    attribute: $2
- match: 'full.name.*'
  name: metric_two
  labels:
    system: static
    attribute: $1
  `,
			mappings: mappings{
				"whatever.dummy.test": {
					name: "metric_one",
					labels: map[string]string{
						"system": "whatever",
                        "attribute": "test",
					},
				},
				"full.name.anothertest": {
					name: "metric_two",
					labels: map[string]string{
						"system": "static",
                        "attribute": "anothertest",
					},
				},
			},
		},
```
Also, I think there is something wrong with the scenario starting with the comment 

> //Config with backtracking

Where the label of the second metric should either have produced an error or be the literal `${1}_foo` because there is no `*`-parts to the metric.

A workaround for me is to set `glob_disable_ordering` to true, but I'm not sure if that has any other implications on our mappings - as we have quite a lot of them.

fffonion:
In some use case that the mapping rules need backtracking and the
non-matched rule has star(s), the fsm will incorrectly return the
captures from latter non-matched rule.

This PR fixes this by copying the current `captures` slice to a different
slice and only return the copied slice. So that later assignment to
`captures` won't affect the captures of a previously matched rule.

Fix #168

### DIFF
--- a/pkg/mapper/fsm/fsm.go
+++ b/pkg/mapper/fsm/fsm.go
@@ -138,6 +138,7 @@ func (f *FSM) GetMapping(statsdMetric string, statsdMetricType string) (*mapping
 	var finalState *mappingState
 
 	captures := make([]string, len(matchFields))
+	finalCaptures := make([]string, len(matchFields))
 	// keep track of captured group so we don't need to do append() on captures
 	captureIdx := 0
 	filedsCount := len(matchFields)
@@ -193,6 +194,8 @@ func (f *FSM) GetMapping(statsdMetric string, statsdMetricType string) (*mapping
 				} else if finalState == nil || finalState.ResultPriority > state.ResultPriority {
 					// if we care about ordering, try to find a result with highest prioity
 					finalState = state
+					// do a deep copy to preserve current captures
+					copy(finalCaptures, captures)
 				}
 				break
 			}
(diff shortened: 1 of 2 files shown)


matthiasr:
Thank you for the prompt fix!

PR #171, opened 2018-12-17
Bump prometheus/client_golang to v0.9.2

simonpasquier:
The latest Go version (1.11.4) breaks the build process because it changes the checksum of the `github.com/prometheus/procfs` module. Bumping prometheus/client_golang to v0.9.2 solves the issue because it uses Go modules which removes `procfs` as an indirect dependency.

See https://github.com/prometheus/blackbox_exporter/issues/395 and https://github.com/golang/go/issues/29278#issuecomment-447537558 for more details.


matthiasr:
Thank you!

PR #175, opened 2019-01-02
Travis CI: add staticcheck target

### LINKED ISSUE #173: staticcheck / go module issues with golang/circleci:1.11
https://circleci.com/gh/prometheus/statsd_exporter/347#artifacts/containers/0

I can reproduce this locally (on macOS, using https://github.com/dziemba/mobymac) with the golang/circleci:1.11 image but not the golang:1.11.4 image or on Darwin directly.

For now, I'm cutting out staticcheck from the build to unblock the 0.9.0 release candidate.

simonpasquier:
Attempt to close #173

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,14 +11,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# By being first, this takes over plain `make`. It cuts out `staticcheck`
-# because in the circleci/golang:1.11 image it chokes on the go module cache.
-default: precheck style unused build test
-
 include Makefile.common
 
-STATICCHECK_IGNORE = \
-  github.com/prometheus/statsd_exporter/main.go:SA1019 \
+STATICCHECK_IGNORE =
 
 DOCKER_IMAGE_NAME       ?= statsd-exporter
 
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -9,5 +9,5 @@ go:
 go_import_path: github.com/prometheus/statsd_exporter
 
 script:
-- make
+- make precheck style unused test staticcheck build
```
matthiasr: or equivalently

```suggestion
- make all
```

matthiasr:
It still doesn't work on CircleCI (see #176), I'm really curious what the difference is between the two?

simonpasquier:
So having `test` before `staticcheck` makes Travis CI happy. Honestly I have no clue why Travis fails to run `staticcheck` first... I'm not seeing the issue on my local machine neither Circle CI but @discordianfish also reported the problem on IRC.

matthiasr:
I originally encountered this on CircleCI.

a dependency declaration like `foo: one two three` does not actually guarantee ordering – in practice, make usually walks the list left-to-right, but it does not know that two should come before three. If this is what we want to force, add a dependency `staticcheck: test` – that way make knows that it needs to test before staticchecking.

if it's not only a problem for this repo, should we make this change in Makefile.common? I'm happy to do it locally first though.

simonpasquier:
This is now using the latest version of `Makefile.common` that deals with the weirdness we were seeing with `staticcheck` and Go modules. It also enables additional checks that have been fixed along the way.

matthiasr:
Wonderful, thank you!

PR #176, opened 2019-01-02
Bring back staticcheck on CircleCI

matthiasr:
this makes CircleCI use the "all" make target explicitly. In f130698
d712534 I added a new "default" target that does not include it, due to
issue #173.

PR #175 does basically this, but for TravisCI. I like the idea of not
relying on the default target.

Maybe this is fixed in CircleCI now?


matthiasr:
😢

PR #178, opened 2019-01-09
Scale summaries from milliseconds to seconds.

### LINKED ISSUE #177: The scaling of timer data should be unified, or (preferably) user specified.
https://github.com/prometheus/statsd_exporter/blob/2d35b097ae242e0a15317d7dfb5f8be0f199567e/exporter.go#L407 Scales histograms by 1000, but the same isn't applied to summaries? In reality, I prefer that either no scaling occurs, or that we have a yaml entry to allow us to specify scaling of timer data. Short term though, just unification of summary and histograms would be very nice, as it can come as quite a surprise when you switch data between the two while setting up your mappings.

SpencerMalone:
Fixes #177 

EDIT: Oh bleh, I'll actually pull and build and ensure this works, and not rely on CI

- [x] Fix Tests to match new scaling
- [x] Fix commit sign off

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -420,7 +420,7 @@ func (b *Exporter) handleEvent(event Event) {
 				mapping,
 			)
 			if err == nil {
-				summary.Observe(event.Value())
+				summary.Observe(event.Value() / 1000) // prometheus presumes seconds, statsd millisecond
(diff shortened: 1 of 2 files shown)


matthiasr:
I rely on CI a lot 😄 let me know when you're ready for a review!

PR #182, opened 2019-01-29
Adds healthcheck for checking the status of the container

vaibhavkhurana2018:
What this PR does / why we need it:

Adds a healthcheck for checking the status of the metrics endpoint. 

Which issue(s) this PR fixes : This will help us recover in sitiuations like getting Internal Server Error and the container being up. 

@matthiasr Please review this. Thanks!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,5 +4,5 @@ LABEL maintainer="The Prometheus Authors <prometheus-developers@googlegroups.com
 COPY statsd_exporter /bin/statsd_exporter
 
 EXPOSE      9102 9125 9125/udp
-HEALTHCHECK CMD curl --fail http://localhost:9102/metrics || exit 1
+HEALTHCHECK CMD wget --spider -S "http://localhost:9102/metrics" -T 60 2>&1 || exit 1
 ENTRYPOINT  [ "/bin/statsd_exporter" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,4 +4,5 @@ LABEL maintainer="The Prometheus Authors <prometheus-developers@googlegroups.com
 COPY statsd_exporter /bin/statsd_exporter
 
 EXPOSE      9102 9125 9125/udp
+HEALTHCHECK CMD curl --fail http://localhost:9102/metrics || exit 1
 ENTRYPOINT  [ "/bin/statsd_exporter" ]


Dockerfile
```diff
@@ -4,4 +4,5 @@ LABEL maintainer="The Prometheus Authors <prometheus-developers@googlegroups.com
 COPY statsd_exporter /bin/statsd_exporter
 
 EXPOSE      9102 9125 9125/udp
+HEALTHCHECK CMD curl --fail http://localhost:9102/metrics || exit 1
```
matthiasr: Unfortunately `curl` is not available in the image, maybe this works?

```suggestion
HEALTHCHECK CMD wget --spider http://localhost:9102/metrics || exit 1
```

(this is busybox wget)

vaibhavkhurana2018: This will not work, as it only checks the availability of the page, and statsd redirect it back to the default page. Will check and update the best possible way for this.

matthiasr: Did you find a way?

vaibhavkhurana2018: Added the timeout to the command, this actually works. Not sure how to reproduce the issue but, i faced the same today and this worked. 
```
vaibhav ~ $ sudo docker ps
CONTAINER ID        IMAGE                         COMMAND                  CREATED             STATUS              PORTS               NAMES
1baa9f5abb44        prom/statsd-exporter:v0.6.0   "/bin/statsd_exporte…"   5 hours ago         Up 5 hours                              statsd-exporter
vaibhav ~ $ sudo docker exec -it 1baa9f5abb44 sh
/ # wget --spider -S "http://localhost:9102/metrics" -T 60 2>&1
Connecting to localhost:9102 (127.0.0.1:9102)
  HTTP/1.1 500 Internal Server Error
wget: server returned error: HTTP/1.1 500 Internal Server Error
```

vaibhavkhurana2018: Well i still am not able to find why this is happening, can you suggest some way to debug the root cause as well.

matthiasr: Typically, this means that we are trying to serve an invalid combination of metrics. Please check existing issues, and if there isn't one that matches, open a new issue with

* your mapping configuration
* a sample of inputs that reproduces the issue (debug logging may help see what happens just before it starts failing)
* the full output of the `/metrics` endpoint when it is failing – this usually gives a clue as to what is invalid

PR #184, opened 2019-02-08
Support for set type

### LINKED ISSUE #183: Set type not supported
In statsd, the set type counts unique occurrences between flushes.   This is not supported in statsd_exporter because "between flushes" is pretty meaningless in a pull system, especially if there are multiple servers scraping the exporter.

Originally requested in https://github.com/prometheus/statsd_exporter/issues/112

@matthiasr requested opening a new issue if somebody had a good idea on how to implement Set.   I wouldn't say that I have a good idea, but I do have some thoughts.

Option 1: assume single scraping server.   Not a great solution, but would be sufficient for us, at least at the moment.

Option 2: create a statsd plugin that sends sets as gauges on flush.   Requires the use of the statsd daemon.

Option 3: add an option for the flush interval; create a ticker from it that persists and resets the set counts every tick.   If the option isn't set it could have a default, or it could just mean the user doesn't need set support.

#3 seems the best option, but #2 is definitely easier and good enough for us.   So we'd love to help with #3, but if there's no interest we'd probably just go ahead and do #2 ourselves.

bryanlarsen:
fixes #183 

still todo:

- [ ] flush interval option
- [ ] support strings, not just floats
- [ ] check for leaking goroutines during configuration reload

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -261,6 +261,25 @@ func (t *TimerEvent) Value() float64                { return t.value }
 func (c *TimerEvent) Labels() map[string]string     { return c.labels }
 func (c *TimerEvent) MetricType() mapper.MetricType { return mapper.MetricTypeTimer }
 
+type SetEvent struct {
+	metricName string
+	value      float64
+	labels     map[string]string
+}
+
+func (t *SetEvent) MetricName() string            { return t.metricName }
+func (t *SetEvent) Value() float64                { return t.value }
+func (c *SetEvent) Labels() map[string]string     { return c.labels }
+func (c *SetEvent) MetricType() mapper.MetricType { return mapper.MetricTypeGauge }
+
+type FlushEvent struct {
+}
+
+func (t *FlushEvent) MetricName() string            { return "flush" }
+func (t *FlushEvent) Value() float64                { return 0.0 }
+func (c *FlushEvent) Labels() map[string]string     { return map[string]string{} }
+func (c *FlushEvent) MetricType() mapper.MetricType { return mapper.MetricTypeGauge }
+
 type Events []Event
 
 type LabelValues struct {
@@ -269,12 +288,25 @@ type LabelValues struct {
 	ttl              time.Duration
 }
 
+type SetStore struct {
+	store  map[string]map[float64]bool
+	events map[string]*SetEvent
+}
+
+func NewSetStore() *SetStore {
+	return &SetStore{
+		store:  make(map[string]map[float64]bool),
+		events: make(map[string]*SetEvent),
+	}
+}
+
 type Exporter struct {
 	Counters    *CounterContainer
 	Gauges      *GaugeContainer
 	Summaries   *SummaryContainer
 	Histograms  *HistogramContainer
 	mapper      *mapper.MetricMapper
+	Sets        *SetStore
 	labelValues map[string]map[uint64]*LabelValues
 }
 
@@ -364,6 +396,35 @@ func (b *Exporter) handleEvent(event Event) {
 		} else {
 			log.Debugf(regErrF, metricName, err)
 			conflictingEventStats.WithLabelValues("counter").Inc()
+		}
+
+	case *FlushEvent:
+		log.Debugf("FlushEvent")
+		for k, v := range b.Sets.store {
+			setevent := b.Sets.events[k]
+
+			b.handleEvent(&GaugeEvent{
+				relative:   false,
+				metricName: setevent.MetricName(),
+				value:      float64(len(v)),
+				labels:     setevent.labels,
+			})
+
+			b.Sets.store[k] = map[float64]bool{}
+		}
+
+	case *SetEvent:
+		store, found := b.Sets.store[metricName]
+		if found {
+			_, have := store[event.Value()]
+			if !have {
+				store[event.Value()] = true
+			}
+		} else {
+			b.Sets.store[metricName] = map[float64]bool{
+				event.Value(): true,
+			}
+			b.Sets.events[metricName] = ev
 		}
 
 	case *GaugeEvent:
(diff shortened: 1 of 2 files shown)


exporter.go
```diff
@@ -366,6 +398,35 @@ func (b *Exporter) handleEvent(event Event) {
...
+				metricName: setevent.MetricName(),
+				value:      float64(len(v)),
+				labels:     setevent.labels,
+			})
+
+			b.Sets.store[k] = map[float64]bool{}
+		}
+
+	case *SetEvent:
+		store, found := b.Sets.store[metricName]
```
matthiasr: what happens if the event has dogstatsd tags attached? I think we need to keep separate sets for each of them.

exporter.go
```diff
@@ -366,6 +398,35 @@ func (b *Exporter) handleEvent(event Event) {
 			conflictingEventStats.WithLabelValues("counter").Inc()
 		}
 
+	case *FlushEvent:
```
matthiasr: I'm not sure about handling the flush in-line with incoming statsd events. It's fundamentally different (internal, not user-generated. Would it make sense to break it out into its own loop, or does that get us into locking/mutex hell?

matthiasr: if we keep it inline, please add a comment explaining why it's here

exporter.go
```diff
@@ -366,6 +398,35 @@ func (b *Exporter) handleEvent(event Event) {
...
+			})
+
+			b.Sets.store[k] = map[float64]bool{}
+		}
+
+	case *SetEvent:
+		store, found := b.Sets.store[metricName]
+		if found {
+			_, have := store[event.Value()]
+			if !have {
```
matthiasr: it's customary in Go to collapse these lines to

```
if _, ok := store[event.Value()]; !ok {
	store[event.Value()] = true
}
```

so that it can go out of scope when no longer needed.

matthiasr: I also see the variable usually called `ok` – you're calling it both `have` and `found`, I'd rather be consistent with those where possible.

exporter.go
```diff
@@ -366,6 +398,35 @@ func (b *Exporter) handleEvent(event Event) {
 			conflictingEventStats.WithLabelValues("counter").Inc()
 		}
 
+	case *FlushEvent:
+		log.Debugf("FlushEvent")
+		for k, v := range b.Sets.store {
+			setevent := b.Sets.events[k]
+
+			b.handleEvent(&GaugeEvent{
```
matthiasr: won't this cause strange side effects? we allow matching on the metric type, so if I have a mapping that explicitly only applies to sets, it won't affect this event, but others will.

matthiasr: One thing you could to, instead of sending a GaugeEvent through the event handling again, is to maintain and set prometheus Gauge metrics directly, based on the mapping of the set event.

exporter.go
bryanlarsen: not familiar with dogstatsd, but that makes sense.   Are those stored in Event.Labels()?

exporter.go
bryanlarsen: I'm not sure how much locking/mutex work there would be, I'm not super familiar with your code.   This mechanism let me completely avoid the question.  :)

exporter.go
bryanlarsen: Did it that way to avoid aliasing.   Is aliasing generally not frowned on in Go code?

matthiasr: it's probably better not to – although my instinct is that if you run into aliasing, you're nesting the code too deeply. One thing that took me a while to get used to in Go style is that we prefer

```
if … {
  … do stuff …
  return
}
… do other stuff …
```

rather than using an `else`. In most cases this means there is only one "ok" to work with. I don't see how to do that neatly here though, so I'm okay with leaving the names as they are.

exporter.go
matthiasr: you would need to mutex-protect all the map accesses. it's fine this way, but I think we can't safely circle back into `handleEvent` because that would cause double-mapping.

exporter.go
bryanlarsen: I found myself duplicating much of the handleEvent function (to get mapping, help, labels), code which I don't completely understand yet.   Doing it this way let me avoid the duplication.

As far as I'm concerned, a set is a type of gauge; anything you'd want to do to a gauge (mapping, histograms as discussed, et cetera) probably would be useful on a set.   So I also rationalize my laziness that way.   It'd have to be called out in the documentation if so.

bryanlarsen:
> Please also remember to add tests for set events, including mappings with `match_type: set`.

I did a quick look, but I didn't find a test to copy & modify to get me started.   Can you point me at one?

exporter.go
matthiasr: yes

exporter.go
matthiasr: but these have all been already determined when mapping the set event. the mappings act on statsd types, not Prometheus types, so even if the output is a gauge we cannot treat `|s` events as the same.

I think the correct way is to preserve whatever is needed from the first event that is handled, and then get the Prometheus metric objects the same way `case *GaugeEvent` does.

matthiasr:
There is no test to copy&paste because we don't have this kind of time-delayed behavior yet. At the moment, we test two pieces separately: [the mapping](https://github.com/prometheus/statsd_exporter/blob/master/pkg/mapper/mapper_test.go), and on a higher level the [event handling](https://github.com/prometheus/statsd_exporter/blob/master/bridge_test.

When you teach `pkg/mapper` about set events, add a test for the configuration around that. The existing event handling tests will cover the parsing of the set events. You will have to add a new test that sends several set events, then triggers a flush, verifies that the gauges are correct, sends more set events, flushes again, and verifies the gauge values again. I know this is a bit tedious to write, but if we don't have this test, the functionality will rot away very quickly (or be incorrect to begin with, how would we know?)

matthiasr:
@bryanlarsen are you still interested in working on this?

bryanlarsen:
We're not using this capability at the moment, so I haven't been working on it.  It'd be nice to use it, but at this point you certainly can't count on me to finish this PR.

matthiasr:
Thanks, in that case I'll close it for now – if you or anyone else wants to pick it up, I'm happy to reopen it.

PR #187, opened 2019-02-28
Add linux UDP buffer telemetry using procfs

SpencerMalone:
We've had some problems when we cannot deploy a statsd exporter on localhost. Our exporters had misleading rate data, because the statsd exporter wasn't keeping up with events, and our linux UDP buffer was overflowing. A major part in fixing that is making it easier to track the UDP buffer overflow rate (and while I was at it, the UDP buffer current depth). This PR is for that thing.

Alternatives I explored:
I originally wanted to use https://github.com/google/cadvisor to collect this data, but was scared off of collecting UDP buffer data by a warning on one of their github issues proclaiming... `tcp/udp create an _enormous_ number of additional metric streams compared with the basic metrics`. I didn't want a large amount of metrics, I just wanted a few

I had also hoped to use the procfs library seen in the prom org, but could not find support in that library for these metrics, so I moved away from that.

I excluded non-linux OSes from this data collection because I don't have a great way to test them. 

Very open to suggestions, lemme know your thoughts.

### DIFF
--- a/main.go
+++ b/main.go
@@ -18,6 +18,7 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"runtime"
 	"strconv"
 
 	"github.com/howeyc/fsnotify"
@@ -181,6 +182,10 @@ func main() {
 
 		ul := &StatsDUDPListener{conn: uconn}
 		go ul.Listen(events)
+
+		if runtime.GOOS == "linux" {
+			go watchUDPBuffers(0, 0)
+		}
 	}
 
 	if *statsdListenTCP != "" {
--- a/telemetry_test.go
+++ b/telemetry_test.go
@@ -1,0 +1,41 @@
+// Copyright 2013 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"io/ioutil"
+	"testing"
+)
+
+func TestProcfsWatching(t *testing.T) {
+	filename := "/tmp/procfstest"
+
+	d1 := []byte("  sl  local_address                         remote_address                        st tx_queue rx_queue tr tm->when retrnsmt   uid  timeout inode ref pointer drops\n42429: 00000000000000000000000000000000:1FBD 00000000000000000000000000000000:0000 07 00000000:00000034 00:00000000 00000000     0        0 1233268343 2 ffff881fc5d32ec0 10000\n")
+	err := ioutil.WriteFile(filename, d1, 0644)
+	if err != nil {
+		t.Fatalf("Should be able to write a procfs-like file: %s", err)
+	}
+
+	queued, dropped, err := parseProcfsNetFile(filename)
+	if err != nil {
+		t.Fatalf("Parsing encountered an error: %s", err)
+	}
+	if dropped != 10000 {
+		t.Fatal("Dropped should be 10000")
+	}
+
+	if queued != 52 {
+		t.Fatal("Queued should be 52", queued)
+	}
+}
--- a/telemetry.go
+++ b/telemetry.go
@@ -14,7 +14,15 @@
 package main
 
 import (
+	"bufio"
+	"os"
+	"runtime"
+	"strconv"
+	"strings"
+	"time"
+
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/log"
 )
 
 var (
@@ -102,6 +110,20 @@ var (
 		},
 		[]string{"type"},
 	)
+	udpBufferQueued = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "statsd_exporter_udp_buffer_queued",
+			Help: "The number of queued UDP messages in the linux buffer.",
+		},
+		[]string{"protocol"},
+	)
+	udpBufferDropped = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "statsd_exporter_udp_buffer_dropped",
+			Help: "The number of dropped UDP messages in the linux buffer",
+		},
+		[]string{"protocol"},
+	)
 )
 
 func init() {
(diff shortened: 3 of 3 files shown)


brian-brazil:
This is machine-level monitoring, which is the responsibility of the node exporter. I think you want the UDP InErrors metric, which is already supported.

SpencerMalone:
Hrm, any idea how I could get that data out of a container? I noticed that both `netstat` and the node exporters that are run on the same hosts as these containers have conflicting UDP drop rate information and seem unable to track when UDP buffer overflows happen in the containers themselves. I'm guessing maybe there's some permissions hijinx? Lemme poke around with some of our systems peeps.

SuperQ:
@brian-brazil I need to check for sure, but I think this would be the correct way to do this. I think the kernel is monitoring per-PID stats.

I would like to see this parsing added to the procfs library, as it would be useful for things other than the statsd_exporter.

SpencerMalone:
Lemme spend some more time swimming around in that codebase and I can see about opening a PR there.

brian-brazil:
> I think the kernel is monitoring per-PID stats.

It's only namespaced, not per-PID.

telemetry_test.go
```diff
@@ -0,0 +1,41 @@
...
+func TestProcfsWatching(t *testing.T) {
+	filename := "/tmp/procfstest"
+
+	d1 := []byte("  sl  local_address                         remote_address                        st tx_queue rx_queue tr tm->when retrnsmt   uid  timeout inode ref pointer drops\n42429: 00000000000000000000000000000000:1FBD 00000000000000000000000000000000:0000 07 00000000:00000034 00:00000000 00000000     0        0 1233268343 2 ffff881fc5d32ec0 10000\n")
+	err := ioutil.WriteFile(filename, d1, 0644)
+	if err != nil {
+		t.Fatalf("Should be able to write a procfs-like file: %s", err)
+	}
+
+	queued, dropped, err := parseProcfsNetFile(filename)
```
matthiasr: if you make this function take a reader (or a `[]byte` even), you don't need to write an actual file in unit tests

main.go
```diff
@@ -181,6 +182,10 @@ func main() {
 
 		ul := &StatsDUDPListener{conn: uconn}
 		go ul.Listen(events)
+
+		if runtime.GOOS == "linux" {
```
matthiasr: If at all possible, I would prefer that code that doesn't work on non-linux platforms doesn't get compiled in, not just skipped with an `if`.

matthiasr:
> It's only namespaced, not per-PID

IFF we cannot get any per-socket stats, I would be okay with explaining the limitations of this metric but exposing it anyway. Then users can make their own judgment calls whether there could be anything other that has UDP drops on the machine.

I don't think this can be solved satisfactorily with the node exporter (because there is no way to get the per-namespace stats _if_ the exporter is containerized) or cAdvisor (which would include too much detail from monitoring all sockets on all network namespaces, while we already know exactly which one we're interested in).

SpencerMalone:
> Q: we _know_ the exact listening socket because we create it, can we somehow get stats about that?

I had the same thought poking around last night! We can, we have the technology.

Sounds like I need a todo list. Lemme know if y'all feel that more needs to be on here:

- [ ] Add a procfs change for `net/protocol`, which for now will include `udp`, and `udp6`. More would be nice here, but being honest, I probably don't have the patience to map out every protocol in procfs right now since each one is different.
- [ ] Implement procfs changes in this PR
- [ ] Track only the drop rate of ports we expose
- [ ] Skip compiling udp buffer watching code on non-linux platforms
- [ ] Figure out our error handling scenarios. Maybe bail if we fail to load the files the first time, and after that if we run into any exceptions, wait a few seconds and try again?

matthiasr:
sounds like a good plan! Thank you for taking this on.

brian-brazil:
> because there is no way to get the per-namespace stats if the exporter is containerized

/proc/net/snmp is not namespaced. In addition I'd presume that /proc/net/udp has the same issues as /proc/net/tcp and is `n^2` or worse performance wise.

If you're going to pull in data here it should be from the socket, not anything from /proc that may include other processes.

SpencerMalone:
@brian-brazil - This is a little deeper than I had anticipated going, but if it's a thing, it sounds like a good thing. I think I can read the max buffer size with a syscall using `SO_RCVBUF`, but I am having trouble figuring out how to pull out the dropped packet count. Any pointers on where to head for docs around this?

brian-brazil:
I thought I'd found one previously, but can't find it again now. The kernel has the data anyway.

rtreffer:
For linux it looks like some data is availabile through `sock_diag`: http://man7.org/linux/man-pages/man7/sock_diag.7.html

There should be a way to get the queued data in the receive path (I couldn't easily find it in the golang api). If you know `SO_RCVBUF` and the maximum packet size you'd be able to prove no packet drop in the UDP layer if `queued data + max packet size < maximum buffer size`.

Another simple (but not perfect) option would be to move the kernel queue draining into its own goroutine. That goroutine would read from the socket as quickly as possible and forward all packets to an internal fixed size channel in a non-blocking manner. The main assumption is that you should always be able to read the data fast enough if you do not do any processing on top. The side assumption is that you are probably game over if that does not hold.

You can then exposing the failed channel writes as a metric. This metric would describe a lower bound on dropped packets.

brian-brazil:
> That goroutine would read from the socket as quickly as possible and forward all packets to an internal fixed size channel in a non-blocking manner. 

That's possibly a good idea in any case.

SpencerMalone:
Agreed! We have a wip branch that started with that, and continued through
to processing data in another set of goroutines. Not ready for review yet,
but so far we went from maxing out at 8k events per second to around 165k
events per second with 6 cores (have not tested past that). Excited to
bring that forward some time in the next month.

matthiasr:
That would be a massive improvement! I shied away from it so far because we rely on the implicit serialization from the single processing goroutine a lot – if you're willing to tackle that it would be great!

What do you think about starting this incrementally by first splitting the network-receive from the processing goroutine – that way we would have the metrics to see when processing is too slow? My assumption would be that a plain receive-and-stuff-into-channel will always be faster than doing the actual processing, but if you're still concerned we could also add the network receive buffer depth monitoring that @rtreffer mentioned.

SpencerMalone:
Bleh, I fell off on this. Here's where I got, and where I am going:
I could not for the life of me get a good UDP linux buffer overflow count without procfs. I poked at using netlink / sock_diag, got that working, and can get `queued data` as @rtreffer outlined, but I'm not sure that knowing "some data was lost" is enough. 

I personally strongly prefer being able to say "less that 0.05% of UDP messages received are dropped". If we only have the vague knowledge of "some data was lost", we can only say "we got all the data" or "we lost an unknown amount of data", and the in between is impossible, which pushes people towards setting unmaintainable goals such as "never drop UDP packets". Furthermore, I think you would either have to do that netlink check before you read data each time or accept that you may have untracked buffer drops (because if you did it on a timer, the read operations between the data collection may drop the data enough that you go from a state of "being at max buffer" -> "being under max buffer"). Please correct me if I'm wrong here.

In this case, my preference is definitely accuracy over performance, and we're not looking at slamming procfs constantly. With those in mind, I'm gonna head down the procfs path and put this behind an optional command line flag? If there ends up being cases where the performance is unbearable, we could revisit, but worst case they could just disable the logic for now.

matthiasr:
I _think_ in this case aiming for 0 dropped packets is not entirely unreasonable, since drops on the network path cannot be considered for this count anyway.

I understood the sock_diag approach to include checking before every read. If we combine this with the separate reader goroutine, we can

* prove that we did not drop any packets before they hit user space
* keep track of the ratio of handled vs. "dropped" (by overflowing the go channel buffer) packages

I think this is going to give us more information than looking at (netns-)global drop counts, because will usually include something else.

You said you have a version with the separate read goroutine, and working code for the netlink approach. Could you submit these two (combined is fine, since the code is going to be intertwined)? I think with these two we are in a much better position for everyone, and it doesn't require anyone to make a decision or be confronted with a non-process-scoped metric.

If you think it's still necessary then, I would be willing to accept an optional procfs check as well, but if the above gives enough insight I would prefer not to add this complexity.

SpencerMalone:
Closing for https://github.com/prometheus/statsd_exporter/pull/196

PR #192, opened 2019-03-23
Fixes #191: crash on empty metric name

### LINKED ISSUE #191: Crash on empty metric name
Break-out from https://github.com/prometheus/statsd_exporter/issues/190#issuecomment-474331791

https://github.com/prometheus/statsd_exporter/blob/1036439f69a70064532a1273047cf7f10712daaa/exporter.go#L283 does not check that the metric name is not empty.

With the following mapping:

```
mappings:
- match: .*_bar
  match_type: regex
  name: "${1}"
```

and any message that matches this mapping:

```
echo "foo_bar:1|c" | nc localhost 9125
```

the exporter crashes:

```
INFO[0000] Starting StatsD -> Prometheus Exporter (version=, branch=, revision=)  source="main.go:158"
INFO[0000] Build context (go=go1.12, user=, date=)       source="main.go:159"
INFO[0000] Accepting StatsD Traffic: UDP :9125, TCP :9125  source="main.go:160"
INFO[0000] Accepting Prometheus Requests on :9102        source="main.go:161"
panic: runtime error: index out of range

goroutine 1 [running]:
main.escapeMetricName(0x0, 0x0, 0x7, 0x1558faf)
	/Users/mr/src/github.com/prometheus/statsd_exporter/exporter.go:283 +0xcc
main.(*Exporter).handleEvent(0xc0000e3f28, 0x160d260, 0xc000129f00)
	/Users/mr/src/github.com/prometheus/statsd_exporter/exporter.go:336 +0x1b7
main.(*Exporter).Listen(0xc0000e3f28, 0xc0001921e0)
	/Users/mr/src/github.com/prometheus/statsd_exporter/exporter.go:308 +0x16e
main.main()
	/Users/mr/src/github.com/prometheus/statsd_exporter/main.go:213 +0x1425
```

We should reject samples that result in an empty metric name and log an error instead of crashing.

Additionally, we may want to reject mappings that reference non-existent capture groups but I'm not sure if that is feasible ahead of time.

ivanizag:
Avoids crashing when escaping empty metric names.
The metrics are ignored and a message is logged. 

@matthiasr

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -334,7 +334,8 @@ func (b *Exporter) handleEvent(event Event) {
 	prometheusLabels := event.Labels()
 	if present {
 		if mapping.Name == "" {
-			log.Debugf("The mapping for match \"%v\" generates an empty metric name.", mapping.Match)
+			log.Debugf("The mapping of '%s' for match '%s' generates an empty metric name", event.MetricName(), mapping.Match)
+			errorEventStats.WithLabelValues("empty_metric_name").Inc()
 			return
 		}
 		metricName = escapeMetricName(mapping.Name)
@@ -352,7 +353,7 @@ func (b *Exporter) handleEvent(event Event) {
 		// will cause the exporter to panic. Instead we will warn and continue to the next event.
 		if event.Value() < 0.0 {
 			log.Debugf("Counter %q is: '%f' (counter must be non-negative value)", metricName, event.Value())
-			eventStats.WithLabelValues("illegal_negative_counter").Inc()
+			errorEventStats.WithLabelValues("illegal_negative_counter").Inc()
 			return
 		}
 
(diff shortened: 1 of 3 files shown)


exporter.go
```diff
@@ -333,6 +333,10 @@ func (b *Exporter) handleEvent(event Event) {
 	metricName := ""
 	prometheusLabels := event.Labels()
 	if present {
+		if mapping.Name == "" {
+			log.Debugf("The mapping for match \"%v\" generates an empty metric name.", mapping.Match)
```
matthiasr: Please also add a metric to track this. On line 255 we are already (mis)using the `eventStats` counter to track negative increments, but I think it would be better if you set up a new metric vector for event errors, and also changed the negative-counter to use it.

matthiasr: Also, please also add `event.MetricName()` to the log line, this will make it much easier to track down what input causes this.

ivanizag:
This should be it. Tests for the telemetry counters could be cleaner but I don't know a way to reset them between tests.

matthiasr:
Wonderful, thank you very much!

PR #194, opened 2019-03-28
Allow support for inconsistent label sets by marking metrics registered as unchecked collectors

### LINKED ISSUE #114: 500 error on metrics endpoint
from time to time, looks like the metrics endpoint dies and start returning 500 errors, and only way to restore it is to restart the pod it's on.
Iogs are empty and I didn't see any options to switch on a more verbose logging.
what can I do in order to provide a more detailed log and help debug it?

vsakhart:
In the current version of the statsd exporter, if a metric with a label set is sent to the statsd exporter with a label set inconsistent with an existing registered metric then that metric gets dropped. e.g `foo{a="1"}` is sent first and then  `foo{b="1"}` is sent but then dropped.

This functionality was officially supported in the Go Prometheus Client via the addition of unchecked collectors. This PR makes it so each metric added is an unchecked collector which will allow metrics with inconsistent label sets not to be dropped. 

I couldn't find a more elegant way to add the metric to the container map without string concat so let me know if there could be a better solution.

The test is also covering a lot of cases within one test case given that the change touches all metrics registration code so let me know if you want me to refactor that at all.

This fixes #114.

@matthiasr


exporter.go
```diff
@@ -87,23 +96,28 @@ func NewCounterContainer() *CounterContainer {
 }
 
 func (c *CounterContainer) Get(metricName string, labels prometheus.Labels, help string) (prometheus.Counter, error) {
-	counterVec, ok := c.Elements[metricName]
+	labelNames := getLabelNames(labels)
+	mapKey := metricName + "," + strings.Join(labelNames, ",")
```
matthiasr: you're doing this in several places, please break it out into a function so that we can be sure it's consistent

matthiasr:
Oh also, please add the DCO sign-off to your commits. It's unfortunately required.

vsakhart:
ping @matthiasr :)

matthiasr:
Sorry, I was on vacation for the last week and am still catching up. I'll review this week.

matthiasr:
Thank you!

PR #197, opened 2019-04-02
Increase escapeMetricName performance

bakins:
While investigating something else,I did some profiling and found that `escapeMetricName` took a large amount of processing time. The regex was the majority of this time.  `escapeMetricName` is called on every event processed. 

The valid set of characters for a metric name is fairly small, so we can optimize this a bit.  I replaced the regex in escapeMetricName with a loop over the runes.  I originally used a `strings.Builder` to build the output, but building the string manually provided to be much faster.

This passes tests, but I am sure I am missing something obvious. This is borderline unreadable as well and needs some comments. I did this on a sick day, so I do not trust myself on this. 

For benchmarks, see https://gist.github.com/bakins/f4d342d1cc93f4911ffd4efa9d93ecab

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -21,7 +21,6 @@ import (
 	"hash/fnv"
 	"io"
 	"net"
-	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -44,8 +43,6 @@ const (
 )
 
 var (
-	illegalCharsRE = regexp.MustCompile(`[^a-zA-Z0-9_]`)
-
 	hash   = fnv.New64a()
 	strBuf bytes.Buffer // Used for hashing.
 	intBuf = make([]byte, 8)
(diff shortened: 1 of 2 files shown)


SpencerMalone:
I've been poking at this a bit, I think it's doing the right thing, and I tried playing with some alternate implementations (such as using `unicode.Is()` and a custom `unicode.RangeTable`), but nothing topped this in performance.

I think with some comments this would be great, and I personally think it's worth keeping the `// Replace all illegal metric chars with underscores.` (or similar) comment in place

matthiasr:
This looks good! I agree, a little commenting would be nice, but I don't think it's less readable than the previous solution.

Why not add the benchmark code (only for the version that is actually there)?

matthiasr:
awesome, thank you!

PR #198, opened 2019-04-03
Adding mapping cache

SpencerMalone:
Mapping is one of the most expensive operations done, but throwing a cache in front of it is an easy way to reduce that pain.

Alongside that, updating the config invokes a lock and wipes the cache, and on k8s if your config is a configmap, it will be regularly synced into the container (even with no changes), at which point you will get needless config reloads (which now wipes the cache, just to be safe). Only reloading if the sha changes is a good way to keep the cache around as long as possible.

Here's the benchmark results:
```
BenchmarkGlob10Rules-12                                 	 5000000	       258 ns/op
BenchmarkGlob10RulesCached-12                           	20000000	        71.8 ns/op
BenchmarkRegex10RulesAverage-12                         	 1000000	      1134 ns/op
BenchmarkRegex10RulesAverageCached-12                   	20000000	        71.9 ns/op
BenchmarkGlob100Rules-12                                	 5000000	       260 ns/op
BenchmarkGlob100RulesCached-12                          	20000000	        72.3 ns/op
BenchmarkGlob100RulesMultipleCaptures-12                	 1000000	      1908 ns/op
BenchmarkGlob100RulesMultipleCapturesCached-12          	20000000	        73.6 ns/op
BenchmarkRegex100RulesMultipleCapturesWorst-12          	  100000	     11116 ns/op
BenchmarkRegex100RulesMultipleCapturesWorstCached-12    	20000000	        74.7 ns/op
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -91,7 +91,7 @@ func tcpAddrFromString(addr string) *net.TCPAddr {
 	}
 }
 
-func watchConfig(fileName string, mapper *mapper.MetricMapper) {
+func watchConfig(fileName string, mapper *mapper.MetricMapper, cacheSize int) {
 	watcher, err := fsnotify.NewWatcher()
 	if err != nil {
 		log.Fatal(err)
@@ -106,7 +106,7 @@ func watchConfig(fileName string, mapper *mapper.MetricMapper) {
 		select {
 		case ev := <-watcher.Event:
 			log.Infof("Config file changed (%s), attempting reload", ev)
-			err = mapper.InitFromFile(fileName)
+			err = mapper.InitFromFile(fileName, cacheSize)
 			if err != nil {
 				log.Errorln("Error reloading config:", err)
 				configLoads.WithLabelValues("failure").Inc()
@@ -149,6 +149,7 @@ func main() {
 		statsdUnixSocketMode = kingpin.Flag("statsd.unixsocket-mode", "The permission mode of the unix socket.").Default("755").String()
 		mappingConfig        = kingpin.Flag("statsd.mapping-config", "Metric mapping configuration file name.").String()
 		readBuffer           = kingpin.Flag("statsd.read-buffer", "Size (in bytes) of the operating system's transmit read buffer associated with the UDP or Unixgram connection. Please make sure the kernel parameters net.core.rmem_max is set to a value greater than the value specified.").Int()
+		cacheSize            = kingpin.Flag("statsd.cache-size", "Maximum size of your metric mapping cache. Relies on least recently used replacement policy if max size is reached.").Default("1000").Int()
 		dumpFSMPath          = kingpin.Flag("debug.dump-fsm", "The path to dump internal FSM generated for glob matching as Dot file.").Default("").String()
 	)
 
@@ -247,7 +248,7 @@ func main() {
 
 	mapper := &mapper.MetricMapper{MappingsCount: mappingsCount}
 	if *mappingConfig != "" {
-		err := mapper.InitFromFile(*mappingConfig)
+		err := mapper.InitFromFile(*mappingConfig, *cacheSize)
 		if err != nil {
 			log.Fatal("Error loading config:", err)
 		}
@@ -257,7 +258,9 @@ func main() {
 				log.Fatal("Error dumping FSM:", err)
 			}
 		}
-		go watchConfig(*mappingConfig, mapper)
+		go watchConfig(*mappingConfig, mapper, *cacheSize)
+	} else {
+		mapper.InitCache(*cacheSize)
 	}
 	exporter := NewExporter(mapper)
 
--- a/pkg/mapper/mapper.go
+++ b/pkg/mapper/mapper.go
@@ -20,6 +20,7 @@ import (
 	"sync"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/log"
 	"github.com/prometheus/statsd_exporter/pkg/mapper/fsm"
 	yaml "gopkg.in/yaml.v2"
 	"time"
@@ -49,7 +50,8 @@ type MetricMapper struct {
 	FSM      *fsm.FSM
 	doFSM    bool
 	doRegex  bool
-	mutex    sync.Mutex
+	cache    MetricMapperCache
+	mutex    sync.RWMutex
 
 	MappingsCount prometheus.Gauge
 }
@@ -83,7 +85,7 @@ var defaultQuantiles = []metricObjective{
 	{Quantile: 0.99, Error: 0.001},
 }
 
-func (m *MetricMapper) InitFromYAMLString(fileContents string) error {
+func (m *MetricMapper) InitFromYAMLString(fileContents string, cacheSize int) error {
 	var n MetricMapper
 
 	if err := yaml.Unmarshal([]byte(fileContents), &n); err != nil {
@@ -189,6 +191,8 @@ func (m *MetricMapper) InitFromYAMLString(fileContents string) error {
 
 	m.Defaults = n.Defaults
 	m.Mappings = n.Mappings
+	m.InitCache(cacheSize)
+
 	if n.doFSM {
 		var mappings []string
 		for _, mapping := range n.Mappings {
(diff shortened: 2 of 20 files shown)


pkg/mapper/mapper.go
```diff
@@ -206,19 +212,39 @@ func (m *MetricMapper) InitFromYAMLString(fileContents string) error {
...
+		}
+		m.sha = newSha
+		return true, nil
+	}
+
+	return false, nil
 }
 
 func (m *MetricMapper) GetMapping(statsdMetric string, statsdMetricType MetricType) (*MetricMapping, prometheus.Labels, bool) {
+	m.mutex.RLock()
```
SpencerMalone: This was moved to the top of `GetMapping` after `-race` pinged me about `doFSM` having a race condition between the reload thread and the Mapping thread.

pkg/mapper/mapper_cache.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+type MetricMapperCacheResult struct {
+	Mapping *MetricMapping
+	Matched bool
+	Labels  prometheus.Labels
+}
+
+type MetricMapperCache struct {
+	cache sync.Map
```
SpencerMalone: I know there's a lot of caveats around this class, and I originally had this implemented through custom map locking behavior for future thread safety, but realize that this perfectly fits the first one of the use cases:
> The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.

bakins: This cache will grow without bounds. We should probably define a maximum number of items we will cache and a cache replacement strategy,

bakins:
There are two unrelated changes in this PR: caching mapping lookups and changing the way config file reloads are handled.  IMO, Should be two PRs - one for each behavioral change.

pkg/mapper/mapper_cache.go
SpencerMalone: Any thoughts on good patterns or resources to look at for cache replacement strategies? It’s not something I am super familiar with. My original thought had been that it would be a cache with a natural limit because there should be an upper limit on the cardinality  you are pushing into the service (since the service ingesting it has a natural limit), but I think you’re right that people could be using this to take high cardinality data and reduce the cardinality before exposing it to prometheus.

My only other thought would be is it adequate to be able to opt-out of the cache on a per-metric basis? IE, if I know this is something that will have a high cardinality on the incoming event level, explicitly don’t cache it?

SpencerMalone:
I think that's a fair point, I went down this path after some prior PRs where it sounded like the preference was to group like-changes together, but the context was different (grouping together a behavioral change and an observability increase VS grouping together two behavioral changes). I'll split off the cache reloading into a new PR

pkg/mapper/mapper_cache.go
bakins: I've used this in the past: https://github.com/hashicorp/golang-lru - you could have the default size be 0, which could mean no cache.

Opting in/out on a per metric basis is interesting, but I think it may complicate both the code and the user experience.

SpencerMalone: Cool! Lemme give this a poke and see how the perf holds up. Will take ~ a week or so.

matthiasr: To what extent will it grow? By design, the exporter consumes some memory per time series, so if cleaning up this cache outside of configuration reloads is difficult I would be okay with not being very clever about it until there is evidence of actual issues.

bakins: @SpencerMalone mentioned to me that we have seen memory issues in our fork of statsd-exporter which implements this cache. We will do some testing to see if the cache is the culprit. For most people, it will probably be fine, but we have a few unusual circumstances in our environment.

matthiasr: makes sense, I'm not opposed to an LRU or similar mechanism if it's worth it!

SpencerMalone:
So, using the LRU cache brought me from 70 ns/op -> ~118 ns/op, which to me is acceptable for the feature gain of a max size, but adding a hit/miss counter bumped me up to 500 ns/op from a goroutine, and to ~220 ns/op for a single threaded approach, neither of which seemed worth it to me.

With that in mind, I'll be tracking the size of the cache, but not a counter of cache requests.

main.go
```diff
@@ -144,6 +144,8 @@ func main() {
 		mappingConfig   = kingpin.Flag("statsd.mapping-config", "Metric mapping configuration file name.").String()
 		readBuffer      = kingpin.Flag("statsd.read-buffer", "Size (in bytes) of the operating system's transmit read buffer associated with the UDP connection. Please make sure the kernel parameters net.core.rmem_max is set to a value greater than the value specified.").Int()
 		dumpFSMPath     = kingpin.Flag("debug.dump-fsm", "The path to dump internal FSM generated for glob matching as Dot file.").Default("").String()
+
+		cacheSize = kingpin.Flag("statsd.cache-size", "Maximum size of your metric mapping cache. Relies on least recently used replacement policy if max size is reached.").Default("1000").Int()
```
matthiasr: I would keep the LRU out of the usage and document it in the README maybe?

matthiasr: the unit is not immediately clear from this – "size" suggests bytes but I think it's the number of cached mappings?

```suggestion
		cacheSize = kingpin.Flag("statsd.cache-size", "Maximum number of cached mappings.").Default("1000").Int()
```

pkg/mapper/mapper.go
```diff
@@ -206,19 +211,42 @@ func (m *MetricMapper) InitFromYAMLString(fileContents string) error {
...
+}
+
+func (m *MetricMapper) InitCache(cacheSize int) {
+	m.useCache = true
+	if cacheSize == 0 {
+		m.useCache = false
+	} else {
+		cache, err := NewMetricMapperCache(cacheSize)
+		if err != nil {
+			log.Warnf("Unable to setup metric cache. Performance may be negatively impacted. Caused by: %s", err)
```
matthiasr: Following the bread crumbs, the only error that could happen here would be [invalid configuration](https://github.com/hashicorp/golang-lru/blob/master/simplelru/lru.go#L28). I would prefer to

* prevent negative cache sizes ourselves
* if we fail to initialize the cache for another reason, fail hard

I'm not a fan of silent failures, even if it's something "optional" like a cache.

matthiasr: Aside from that, this is purely a matter of taste (so no need to change it if you disagree): I'm not a fan of "if use $feature …" – could we install a NOOP cache instead?

pkg/mapper/mapper_cache.go
```diff
@@ -0,0 +1,73 @@
...
+	Matched bool
+	Labels  prometheus.Labels
+}
+
+type MetricMapperCache struct {
+	cache *lru.Cache
+}
+
+func NewMetricMapperCache(size int) (*MetricMapperCache, error) {
+	cacheLength.Set(0)
```
matthiasr: is the cacheLength global or per cache? It seems that everything is set up to accommodate multiple metric mapper caches, except for the cacheLength gauge.

I would propose having a prometheus.Gauge passed in by the caller instead?

main.go
SpencerMalone: You've got it!

pkg/mapper/mapper_cache.go
SpencerMalone: It's global, I don't expect to have multiple metric mapper caches, I guess there's nothing stopping you, but I'm not 100% sure what you gain, since the cache implementation is thread safe. I guess decreased lock contention, but in my production test of this with a more threaded backend focusing into one instance of the cache (we've had it running with a cache ~1/2 the size of the incoming metrics to force a high churn), lock contention has not been a problem. I'm not against doing this, but I kind of feel like it's something we can swap to later if we decide it's a path we wanna head down?

pkg/mapper/mapper.go
SpencerMalone: Happy to do so! I should have followed the breadcrumbs more carefully myself.

pkg/mapper/mapper_cache.go
matthiasr: My concern is less about what we need right now, and more about the package interface being confusing – however I just realized that we already have a metric like that, and you're right that it's not something we need to fix right now, so I'm okay with it as it is.

SpencerMalone:
I think this should be good to go!

matthiasr:
Beautiful, thanks a lot for pushing this through!

Issue #189, opened 2019-03-11
Support for Unix domain sockets interface


mohag:
Statsd [supports a Unix domain socket interface](https://github.com/statsd/statsd/pull/564) (which can be used between containers with volumes).

Kong seems to have [added it to their version of statsd_exporter](https://docs.konghq.com/enterprise/0.34-x/vitals-prometheus-strategy/#statsd-exporter-with-unix-domain-socket) as well, but [their repo](https://github.com/kong/statsd_exporter) does not seem to contain the relevant code.

[DogStatsD also supports](https://docs.datadoghq.com/developers/dogstatsd/unix_socket/) this.

This would be nice to have in statsd_exporter as well. Currently the obvious way to do this involves running a statsd instance to feed the stats into statsd_exporter from a Unix domain socket. If this is implemented, compatibility with Kong's options for specifying the socket details might make sense.

matthiasr:
This is a reasonable feature request – the easiest way would be if Kong were to submit their code. @subnetmarco is this something you would be willing to contribute?

fffonion:
The reason for Kong's private fork is mainly to add the feature of "min/max" gauge, which is a stateful gauge that reset after scrape, to support Kong Vitals functionality. And the "reset" pattern is obviously anti Prometheus pattern, so we decided not push it to upstream. (Correct me if I'm wrong : ) )

The work of unix domain socket is done together with the performance research, which incubated the fsm mapper that got merged in upstream. Thanks @matthiasr for a detailed review! The domain socket itself, after some experienment, we found it not gaining notible performance increase. So it's not given much attention. Since now there's a need for that functionality, I'll organize a PR.

SuperQ:
@fffonion Yes, the "reset" pattern is very much not something possible with Prometheus as it breaks badly if you have more than one Prometheus. Having multiple Prometheus for HA is a standard deployment pattern.

If you wanted a min/max, it's best to track all samples of a metric over a specific time period and decay them from the ring buffer. Similar to how the Prometheus Summary is implemented[0].

[0]: https://github.com/prometheus/client_golang/blob/master/prometheus/summary.go

matthiasr:
@fffonion thank you for the context! I agree that min/max is not something to upstream.

@mohag considering the unix domain socket interface does not help with performance, do you think it's still necessary? What is your use case?

SuperQ:
@matthiasr min/max might be an interesting feature proposal for `Summary` type.

[...]

PR #199, opened 2019-04-05
Implement listener for Unixgram sockets

fffonion:
As requested in #189

### DIFF
--- a/main.go
+++ b/main.go
@@ -18,7 +18,9 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"os/signal"
 	"strconv"
+	"syscall"
 
 	"github.com/howeyc/fsnotify"
 	"github.com/prometheus/client_golang/prometheus"
@@ -199,7 +201,8 @@ func main() {
 	}
 
 	if *statsdListenUnixgram != "" {
-		if _, err := os.Stat(*statsdListenUnixgram); !os.IsNotExist(err) {
+		var err error
+		if _, err = os.Stat(*statsdListenUnixgram); !os.IsNotExist(err) {
 			log.Fatalf("Unixgram socket \"%s\" already exists", *statsdListenUnixgram)
 		}
 		uxgconn, err := net.ListenUnixgram("unixgram", &net.UnixAddr{
@@ -225,6 +228,8 @@ func main() {
 		// if it's an abstract unix domain socket, it won't exist on fs
 		// so we can't chmod it either
 		if _, err := os.Stat(*statsdListenUnixgram); !os.IsNotExist(err) {
+			defer os.Remove(*statsdListenUnixgram)
+
 			// convert the string to octet
 			perm, err := strconv.ParseInt("0"+string(*statsdUnixSocketMode), 8, 32)
 			if err != nil {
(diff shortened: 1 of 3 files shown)


fffonion:
There's a question pending dicussion:

If the unix domain socket is not an abstract socket, that means there'll be a file created in the fs. And normally that file will be removed if we call `theSocket.Close()` in go code. But there's currently no signal handling in the exporter, that means the defer logic will not be executed properly and the socket file will not be deleted. Would it be good idea to add a signal handler in this PR so we can clean the socket file or does it worth another PR?

main.go
```diff
@@ -137,27 +137,30 @@ func dumpFSM(mapper *mapper.MetricMapper, dumpFilename string) error {
...
-		dumpFSMPath     = kingpin.Flag("debug.dump-fsm", "The path to dump internal FSM generated for glob matching as Dot file.").Default("").String()
+		listenAddress        = kingpin.Flag("web.listen-address", "The address on which to expose the web interface and generated Prometheus metrics.").Default(":9102").String()
+		metricsEndpoint      = kingpin.Flag("web.telemetry-path", "Path under which to expose metrics.").Default("/metrics").String()
+		statsdListenUDP      = kingpin.Flag("statsd.listen-udp", "The UDP address on which to receive statsd metric lines. \"\" disables it.").Default(":9125").String()
+		statsdListenTCP      = kingpin.Flag("statsd.listen-tcp", "The TCP address on which to receive statsd metric lines. \"\" disables it.").Default(":9125").String()
+		statsdListenUnixgram = kingpin.Flag("statsd.listen-unixgram", "The Unixgram socket path to receive statsd metric lines in datagram. \"\" disables it.").Default("").String()
+		// not using Int here because flag diplays default in decimal, 0755 will show as 493
+		statsdUnixSocketUmask = kingpin.Flag("statsd.unixsocket-umask", "The umask of the unix socket.").Default("755").String()
```
matthiasr: I think umask is the complement of this? Does this _need_ to be configurable, or could we rely on `umask` being run beforehand if someone wants to limit it?

matthiasr:
```suggestion
		statsdUnixSocketMode = kingpin.Flag("statsd.unixsocket-mode", "The permission mode of the unix socket.").Default("755").String()
```

main.go
```diff
@@ -195,6 +198,47 @@ func main() {
...
+		}
+
+		ul := &StatsDUnixgramListener{conn: uxgconn}
+		go ul.Listen(events)
+
+		// if it's an abstract unix domain socket, it won't exist on fs
+		// so we can't chmod it either
+		if _, err := os.Stat(*statsdListenUnixgram); !os.IsNotExist(err) {
+			// convert the string to octet
+			perm, err := strconv.ParseInt("0"+string(*statsdUnixSocketUmask), 8, 32)
```
matthiasr:
```suggestion
			perm, err := strconv.ParseInt("0"+string(*statsdUnixSocketMode), 8, 32)
```

main.go
```diff
@@ -195,6 +198,47 @@ func main() {
...
+			if err != nil {
+				log.Warnf("Bad permission %s: %v, ignoring\n", *statsdUnixSocketUmask, err)
```
matthiasr:
```suggestion
				log.Warnf("Bad permission %s: %v, ignoring\n", *statsdUnixSocketMode, err)
```

main.go
fffonion: Accepted the name suggestions. Regarding if it's needed: i feel it would be helpful to let the process to able to control the file it created, but I'm open to opinions if we should remove it.

fffonion:
@matthiasr 👍 Will add the signal handler in next commits.

matthiasr:
Thank you!

PR #200, opened 2019-04-10
Add metric to track unique metric names the exporter is tracking

bakins:
Currently, we track the number of mappings, but not the unique metricNames.  This adds a simple counter to track the count of metricName+labelNames.  If this number increases significantly, then the application sending metrics may be misbehaving.  This does not count unique label values.

I didn't see a way to get this in the current metrics, so please let me know if there is an existing way to get this information.

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -102,6 +102,7 @@ func (c *CounterContainer) Get(metricName string, labels prometheus.Labels, help
 
 	counterVec, ok := c.Elements[mapKey]
 	if !ok {
+		metricsCount.Inc()
 		counterVec = prometheus.NewCounterVec(prometheus.CounterOpts{
 			Name: metricName,
 			Help: help,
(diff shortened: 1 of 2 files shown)


matthiasr:
Would it be valuable to break this down by type (gauge/counter/summary/histogram)?

SpencerMalone:
I definitely think that is a good idea, since there are performance implications to the two timer types at high cardinality.

bakins:
I can add a label per type, fix the decrement, and make it a gauge.  I was thinking of a slightly different approach as well.  This was a quick PR to gauge interest.

matthiasr:
I think this is a useful thing to track!

bakins:
@matthiasr done. Sorry, I keep forgetting to do that.

matthiasr:
I use an [alias](https://github.com/matthiasr/profile/blob/79cbd47af5db3569e312e89c19c9df17001656ad/gitconfig#L35) so I don't forget 😅

PR #205, opened 2019-04-22
Add pprof for profiling

claytono:
This PR adds pprof and will expose it via the default HTTP endpoint.  This is already in place for other exporter such as [`node_exporter`](https://github.com/prometheus/node_exporter/blob/ffefc8e74dd59143f93af4cecbbae4799bce8c35/node_exporter.go#L19) and [`push_gateway`](https://github.com/prometheus/pushgateway/blob/6b81e91710e704e9f472c1b71f337522c89d03d7/main.go#L19).


matthiasr:
Thank you!

PR #211, opened 2019-05-13
Add support for Datadog distribution type

claytono:
This adds the ability to accept Datadog's distribution type and treat it like a histogram/summary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,8 @@ mappings:
     job: "${1}_server"
 ```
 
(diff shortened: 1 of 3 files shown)


matthiasr:
🙇

gaizeror:
@claytono So when configuring distribution at the mapping yaml file, I should set:
`timer_type: historgram`?

claytono:
@gaizeror Distribution metrics will be treated the same as histogram and timing metrics, so summary by default, and histogram if you change it in the config.

PR #212, opened 2019-05-14
Add default quantiles for summaries if no mapping file exists

### LINKED ISSUE #195: Default timer type is undefined
According to the readme the default timer type is a `summary`. But starting the `statsd_exporter` (version: 0.9.0) without any `statsd.mapping-config` creates just the `_sum` and `_count` time series for a `timer`.

#### Steps to reproduce

* `./statsd_exporter`
* `echo "foo:1234|ms|#one:1,#two:2" | nc -u localhost 9125`
* Check http://localhost:9102/metrics

#### Output
```
foo_sum{one="1",two="2"} 1.234
foo_count{one="1",two="2"} 1
```

#### Expected output
```
foo{one="1",two="2",quantile="0.5"} 1.234
foo{one="1",two="2",quantile="0.9"} 1.234
foo{one="1",two="2",quantile="0.99"} 1.234
foo_sum{one="1",two="2"} 1.234
foo_count{one="1",two="2"} 1
```

I get the expected output by specifying a minimal `mapping-config`: 
```yml
defaults:
  timer_type: summary
```

vsakhart:
Currently the exporter will only report sum/count metrics for summaries if a mapping file isn't defined rather than a default set of quantiles.

This fixes #195.

@matthiasr

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -189,6 +189,10 @@ func (c *SummaryContainer) Get(metricName string, labels prometheus.Labels, help
 		for _, q := range quantiles {
 			objectives[q.Quantile] = q.Error
 		}
+		// In the case of no mapping file, explicitly define the default quantiles
+		if len(objectives) == 0 {
+			objectives = map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001}
+		}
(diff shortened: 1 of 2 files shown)


matthiasr:
Thank you!

Issue #186, opened 2019-02-28
Reject invalid metrics/mappings early instead of breaking `/metrics`


amithad:
When a user submits an invalid metric (such as mapping a counter and a gauge to the same metric name), we do not log anything, but from that point on (until the exporter is restarted) `/metrics` will return a 500. We ought to not accept the second (then invalid) sample at all, and log a descriptive error message instead. This way users do not lose visibility into their other metrics.

Original issue:
=======================

Statsd exporter results an internal error (Showed in Prometheus) when malformed statsd lines are sent. However, this does not show any error in the log as well.

Steps to recreate:

1. Use the statsd package 'hot-shots' (nodejs)
2. Send an 'increment' without specifying the increment number.

`
import {StatsD} from 'hot-shots';
let client  = new StatsD(this.initConfig);
client.increment('adapter.' + adapter + '.errors'); //note the buggy invocation
`
On Prometheus targets, exporter goes to down state. No logs will be shown.

matthiasr:
Is there anything logged with `--log.level=debug`?
Is this a regression in the latest release, or did it happen before?
What is the line that hot-shots sends to the exporter?

matthiasr:
Thank you for reporting! This is definitely an issue, we should reject this line immediately and not poison the Prometheus client.

amithad:
I tried recreating it with a single stat line, didn't happen. However, collectively with all stats being published,

matthiasr:
What is the error message returned with the 500?

amithad:
Error message : server returned HTTP status 500 Internal Server Error

amithad:
However, statsd-exporter docker container seems to continue without exit

amithad:
@matthiasr  I shall clone the repo and run the exporter on debug mode and collect logs for you if you need further information about the bug. Let me know

matthiasr:
Yes, please do! Any additional information helps.

What version of the exporter are you running? What version is running in
the Docker container? When you check out the repo, what happens with the
latest code?

amithad:
Here's the exact scenario;

I publish two stats from my nodejs application using `hotshots`. 
    1. Stat 1 - increment
    2. Stat 2 - gauge

tcpdump of the stat lines sent to statsd-exporter are as follows:

```14:57:14.586530 IP localhost.38727 > localhost.9125: UDP, length 51
E..O$.@.@............G#..;.NMYAPP.Integrator.HEADLESS.adapter.Adapter.errors:1|c

[...]

PR #213, opened 2019-05-14
Add checking for conflicting metrics

claytono:
This adds sanity checking before registering a metric to ensure that the metric isn't already registered under an existing name.  This prevents the exporter from getting into a state where it has accepted and registered conflicting metrics, and then cannot report the metrics it has collected.

I believe this should address the issues raised in #186

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -57,6 +57,19 @@ func (u uncheckedCollector) Collect(c chan<- prometheus.Metric) {
 	u.c.Collect(c)
 }
 
+type metricType int
+
+const (
+	CounterMetricType metricType = iota
+	GaugeMetricType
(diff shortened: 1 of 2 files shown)


claytono:
@matthiasr That's a good idea.  I had problems with labels with my first approach to that, a test case or two to validate that scenario makes sense.

claytono:
@matthiasr Test cases updated and passing CI.  If you can take another  👀 when you have time, I'd appreciate it.

matthiasr:
Awesome, thank you!

claytono:
@matthiasr Thanks for all the fast reviews.  It's greatly appreciated!  👍

PR #216, opened 2019-05-15
Check for histogram conflict on main name

claytono:
While it doesn't report it as a metric, when collecting, the registry considers the base metric name to belong to the histogram that's registered.  This means that we need to prevent other metrics from
registering anything under this name, in addition to checking for the `_bucket`, `_sum` and `_count` suffixes.

This was missed in #213


matthiasr:
Good catch!

PR #217, opened 2019-05-16
Convert escapeMetricName to use strings.Builder

claytono:
This converts `escapeMetricName` to use `strings.Builder` instead of allocating a byte array, filling it and then converting it to a string. This also optimizes for the case where the metricName is already valid, and in that case, it just returns the original string.

This reduces memory allocations from 2 per call to 1 per call in the case when the string does need to be escaped, and reduces it to zero memory allocations when the string is already valid.

As discussed in #197, `escapeMetricName` is called a *lot*, and even after the improvements there, we still see this pretty far up the list in terms of cpu usage in profiling.

Benchmark before:
```
$ go test -benchmem -run=^$ github.com/prometheus/statsd_exporter -bench '^(BenchmarkEscapeMetricName)$' -benchtime 30s
goos: darwin
goarch: amd64
BenchmarkEscapeMetricName/clean-8                       2000000000              33.6 ns/op            10 B/op          2 allocs/op
BenchmarkEscapeMetricName/0starts_with_digit-8          500000000               85.9 ns/op            64 B/op          2 allocs/op
BenchmarkEscapeMetricName/with_underscore-8             1000000000              55.7 ns/op            32 B/op          2 allocs/op
BenchmarkEscapeMetricName/with.dot-8                    1000000000              40.2 ns/op            16 B/op          2 allocs/op
BenchmarkEscapeMetricName/with😱emoji-8                 1000000000              57.4 ns/op            32 B/op          2 allocs/op
BenchmarkEscapeMetricName/with.*.multiple-8             1000000000              58.5 ns/op            32 B/op          2 allocs/op
BenchmarkEscapeMetricName/test.web-server.foo.bar-8     500000000               74.3 ns/op            64 B/op          2 allocs/op
BenchmarkEscapeMetricName/#00-8                         5000000000               9.59 ns/op            0 B/op          0 allocs/op
PASS
ok      github.com/prometheus/statsd_exporter   449.792s
```

Benchmark after:
```
go test -benchmem -run=^$ github.com/prometheus/statsd_exporter -bench '^(B
enchmarkEscapeMetricName)$' -benchtime 30s
goos: darwin
goarch: amd64
BenchmarkEscapeMetricName/clean-8                       5000000000              8.42 ns/op            0 B/op          0 allocs/op
BenchmarkEscapeMetricName/0starts_with_digit-8          1000000000              45.0 ns/op            32 B/op          1 allocs/op
BenchmarkEscapeMetricName/with_underscore-8             3000000000              15.1 ns/op             0 B/op          0 allocs/op
BenchmarkEscapeMetricName/with.dot-8                    1000000000              37.3 ns/op             8 B/op          1 allocs/op
BenchmarkEscapeMetricName/with😱emoji-8                 1000000000              50.5 ns/op            16 B/op          1 allocs/op
BenchmarkEscapeMetricName/with.*.multiple-8             1000000000              58.9 ns/op            16 B/op          1 allocs/op
BenchmarkEscapeMetricName/test.web-server.foo.bar-8     500000000               75.3 ns/op            32 B/op          1 allocs/op
BenchmarkEscapeMetricName/#00-8                         10000000000             2.09 ns/op            0 B/op          0 allocs/op
PASS
ok      github.com/prometheus/statsd_exporter   367.326s
```

As you can see from the benchmarks, the common case where no escaping needs to be done is about 3-4x faster.

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -364,32 +364,56 @@ type Exporter struct {
 // Replace invalid characters in the metric name with "_"
 // Valid characters are a-z, A-Z, 0-9, and _
 func escapeMetricName(metricName string) string {
-	// If a metric starts with a digit, prepend an underscore.
-	if len(metricName) > 0 && metricName[0] >= '0' && metricName[0] <= '9' {
-		metricName = "_" + metricName
+	metricLen := len(metricName)
+	if metricLen == 0 {
+		return ""
 	}
 
-	// this is an character replacement method optimized for this limited
+	escaped := false
+	var sb strings.Builder
+	// If a metric starts with a digit, allocate the memory and prepend an
+	// underscore.
+	if metricName[0] >= '0' && metricName[0] <= '9' {
+		escaped = true
+		sb.Grow(metricLen + 1)
+		sb.WriteByte('_')
+	}
+
+	// This is an character replacement method optimized for this limited
 	// use case.  It is much faster than using a regex.
-	out := make([]byte, len(metricName))
-	j := 0
-	for _, c := range metricName {
-		// check if the rune is valid for a metric name
-		// and replace it if it is not.
-		// As only certain ASCII characters are valid in metric names,
-		// we can use a byte.
-		if (c >= 'a' && c <= 'z') ||
-			(c >= 'A' && c <= 'Z') ||
-			(c >= '0' && c <= '9') {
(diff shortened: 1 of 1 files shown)


matthiasr:
TIL about strings.Builder 😄 thank you!

PR #218, opened 2019-05-16
Reduce memory allocations in handlePacket

claytono:
This converts the byte buffer to a string by casting and parses the
string by looping over it instead of calling Split.

This usage of unsafe is taken from the strings.Builder package here:
https://github.com/golang/go/blob/cb5c82bc3df2e95c2ab36555fefb7cd3f334c685/src/strings/builder.go#L47

This also adds a benchmark for testing overall performance of the UDP code path.

Signed-off-by: Clayton O'Neill <claytono@github.com>

### DIFF
--- a/bridge_test.go
+++ b/bridge_test.go
@@ -315,3 +315,54 @@ func TestHandlePacket(t *testing.T) {
 		}
 	}
 }
+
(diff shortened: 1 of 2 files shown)


matthiasr:
Sweet, thank you!

PR #222, opened 2019-05-23
Fix parsing labels from metric names.

jeamland:
The list of label names used in creating a metric is a list of strings
derived from string:string map containing the actual Prometheus labels.
This map was being acquired from the Event and then any labels parsed out
of the metric name are added. Problematically the list of label names was
acquired before any parsed-out labels were added and thus these would be
missing leading to cardinality errors upon updating the metric.

Also adds a test for this behaviour.

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -458,7 +458,6 @@ func (b *Exporter) handleEvent(event Event) {
 
 	metricName := ""
 	prometheusLabels := event.Labels()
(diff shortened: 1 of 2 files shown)


matthiasr:
this looks great, thank you! Could you please add the sign-off so I can merge it?

matthiasr:
@jeamland because of the missing DCO, I merged #224 for now. However, I still want to add the test – could you please rebase and in the process add the sign-off?

PR #224, opened 2019-05-24
Fix inconsistent label cardinality error

ptqa:
Hi.

I've faced issue with latest statsd version: `Failed to update metric "asr_duration". Error: inconsistent label cardinality: expected 0 label values but got 4 in prometheus.Labels`

It happened after this commit https://github.com/prometheus/statsd_exporter/commit/d143398343bc68ba149e51bdd84acf6e679118fd 

Basically we sort prometheus labels (https://github.com/prometheus/statsd_exporter/blob/master/exporter.go#L461) before fill structure (https://github.com/prometheus/statsd_exporter/blob/master/exporter.go#L470)

Fixes issue for me, also rollback to v0.9.0 fixes it.

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -458,7 +458,6 @@ func (b *Exporter) handleEvent(event Event) {
 
 	metricName := ""
 	prometheusLabels := event.Labels()
-	sortedLabelNames := getSortedLabelNames(prometheusLabels)
 	if present {
 		if mapping.Name == "" {
 			log.Debugf("The mapping of '%s' for match '%s' generates an empty metric name", event.MetricName(), mapping.Match)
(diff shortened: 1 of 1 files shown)


matthiasr:
Thank you for finding this! #222 is functionally the same fix, and it also has a test to prevent regressions, so I hope you don't mind if I close this PR and go with that?

matthiasr:
Change of plan, since #222 is blocked on the DCO sign-off I'm going to merge this for now.

claytono:
Sorry for introducing this.  We don't use label mapping in our environment, so I missed this :(

ptqa:
@claytono no problem, it's a minor thing, happens to everybody

PR #225, opened 2019-05-28
Add a test for adding labels from mapper configuration.

jeamland:
Replaces #222 

Signed-off-by: Benno Rice <benno@jeamland.net>


matthiasr:
thank you!

PR #227, opened 2019-05-30
Add internal event queuing and flushing

claytono:
At high traffic levels, the locking around sending on channels can cause a large amount of blocking and CPU usage.  This adds an event queue mechanism so that events are queued for short period of time, and flushed in batches to the main exporter goroutine periodically.

The default is is to flush every 1000 events, or every 200ms, whichever happens first.

### DIFF
--- a/telemetry.go
+++ b/telemetry.go
@@ -27,7 +27,7 @@ var (
 	)
 	eventsFlushed = prometheus.NewCounter(
 		prometheus.CounterOpts{
-			Name: "statsd_exporter_events_flushed_total",
+			Name: "statsd_exporter_event_queue_flushed_total",
 			Help: "Number of times events were flushed to exporter",
 		},
 	)
--- a/telemetry.go
+++ b/telemetry.go
@@ -25,6 +25,12 @@ var (
 		},
 		[]string{"type"},
 	)
+	eventsFlushed = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Name: "statsd_exporter_events_flushed_total",
+			Help: "Number of times events were flushed to exporter",
+		},
+	)
 	eventsUnmapped = prometheus.NewCounter(prometheus.CounterOpts{
 		Name: "statsd_exporter_events_unmapped_total",
 		Help: "The total number of StatsD events no mapping was found for.",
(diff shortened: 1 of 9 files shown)


matthiasr:
interesting, this somewhat relates to the discussion we had around measuring UDP packet drops here https://github.com/prometheus/statsd_exporter/pull/196#discussion_r273464631

what would happen with the different listeners if the queue overflows?

claytono:
I skimmed through that a while back, and I like the idea in general, but I haven't actually read the PR.  If nothing else, being able to have metrics for dropped packets I think has value.  Another approach to getting those metrics might be to switch the front end listeners to do non-blocking sends on the event queue.  That would allow dropping events when the queue is full and there could be a metric for that.  That might be a simplest change than adding another stage to the processing pipeline.

I think the behavior regarding queue overflows stays mostly the same.  The event queue that the front end listeners push events to now has a bigger buffer, and there is the small internal buffer, but eventually both of those will fill and it will block on trying to send on that channel.

matthiasr:
> Another approach to getting those metrics might be to switch the front end listeners to do non-blocking sends on the event queue

That is my preferred approach as well, because it's platform independent. Does it make sense to add it here?

claytono:
If you're ok with it, I'd prefer to address that in a future PR.  I'm not sure if I'll have the time to incorporate that into this PR in the short term.

telemetry.go
```diff
@@ -25,6 +25,12 @@ var (
 		},
 		[]string{"type"},
 	)
+	eventsFlushed = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Name: "statsd_exporter_events_flushed_total",
```
matthiasr: I would like to make the metric name less ambiguous:

```suggestion
			Name: "statsd_exporter_event_queue_flushed_total",
```

the help makes it clear but is not really visible in Prometheus, so I think it's important to be clear that this is not a count of events but a count of flushes.

claytono:
@matthiasr I think that makes sense.  I'd also like to get that metric in place also, but I'm tied up with another project at the moment.  Your comment about the metric naming makes sense and I've updated the PR to incorporate that.

PR #229, opened 2019-06-06
Make mapper cache respect metric type

### LINKED ISSUE #228: Match metric type not working as documented
## Expected Behavior
2 statsd metrics with the same name but different types (e.g. counter and timer) can be disambiguated with `match_metric_type`.

## Current Behavior
Statsd exporter reports that the metric is already registered when sending a second metric with the same name but different type.

## Steps to Reproduce
1. Use the following mapping:
```
mappings:
- match: test.foo.*
  name: "test_foo"
  match_metric_type: counter
  labels:
    provider: "$1"
- match: test.foo.*
  name: "test_foo_timer"
  match_metric_type: timer
  labels:
    provider: "$1"
```
2. Run statsd exporter (I used Docker) with the mapping and debug logging enabled
3. `echo "test.foo.example:0.003666|ms" | nc -w 1 -u 0.0.0.0 9125`
4. `echo "test.foo.example:1|c" | nc -w 1 -u 0.0.0.0 9125`
5. Step 4 produces the following log line: `time="2019-06-05T12:45:07Z" level=debug msg="Failed to update metric \"test_foo_timer\". Error: metric with name test_foo_timer is already registered" source="exporter.go:500"` and no metric of type `counter` is visible in `/metrics`

## Context (Environment)
We are currently emitting metrics for some tasks in both a timer and a counter format so we can count the number of tasks and time each one. To avoid having to change all this metrics code to use unique names for each type, it would be useful if statsd exporter could work out the type from the statsd packet as per docs for `match_metric_type`.

I've tried dumping the FSM dot file but I'm not hugely familiar with Go or this codebase so not sure exactly what I'm looking for/at. The issue seems to be that the glob expression I am trying to match is repeated but I thought I could disambiguate it using the `match_metric_type`. The FSM dot diagram shows that the glob pattern is registered on both the `counter` and the `timer` branches but the actual behaviour seems to suggest that the statsd metric name is being matched against a glob pattern, before the type is evaluated. Additionally, FSM warns at startup that `time="2019-06-05T12:44:55Z" level=warning msg="match \"test.foo.*\" is a super set of match \"test.foo.*\" but in a lower order, the first will never be matched" source="fsm.go:294"` - even though the type should disambiguate.

Please let me know if you need any more details.

andy-paine:
- Statsd allows users to provide a metric with the same name but
  differing types (counter, gauge, timer)
- The exporter allows this by letting users specify a
  "match_metric_type" in the mapping config
- However the mapper cache does not look at the metric type so it would
  return a MetricMapperCacheResult for the type of the first metric with
that name that the exporter saw
- Add MetricType to the signature for the cache and format the metric
  name with the type to provide unique keys for a metric with the same
name but differing type

Fixes #228

### DIFF
--- a/pkg/mapper/mapper.go
+++ b/pkg/mapper/mapper.go
@@ -237,7 +237,7 @@ func (m *MetricMapper) InitCache(cacheSize int) {
 func (m *MetricMapper) GetMapping(statsdMetric string, statsdMetricType MetricType) (*MetricMapping, prometheus.Labels, bool) {
 	m.mutex.RLock()
 	defer m.mutex.RUnlock()
-	result, cached := m.cache.Get(statsdMetric)
+	result, cached := m.cache.Get(statsdMetric, statsdMetricType)
 	if cached {
 		return result.Mapping, result.Labels, result.Matched
 	}
@@ -253,12 +253,12 @@ func (m *MetricMapper) GetMapping(statsdMetric string, statsdMetricType MetricTy
 				labels[result.labelKeys[index]] = formatter.Format(captures)
 			}
 
-			m.cache.AddMatch(statsdMetric, result, labels)
+			m.cache.AddMatch(statsdMetric, statsdMetricType, result, labels)
 
 			return result, labels, true
 		} else if !m.doRegex {
 			// if there's no regex match type, return immediately
-			m.cache.AddMiss(statsdMetric)
+			m.cache.AddMiss(statsdMetric, statsdMetricType)
 			return nil, nil, false
 		}
 	}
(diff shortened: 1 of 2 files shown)


SpencerMalone:
To me this looks good, do you have thoughts on adding a test? If not, I'm good with this going in, and I can open a followup PR for a test to try and ensure this edge case isn't missed in any future work.

andy-paine:
My Golang isn't all that great - I tried writing some tests but given the lack of tests in general in this repo I didn't really have a pattern to follow and don't feel in a position to be defining patterns for a whole repo. I figured I would raise this now and potentially look to add some tests later.

matthiasr:
ok, I'll merge it then and wait for the test from @SpencerMalone

PR #231, opened 2019-06-08
Add test for multiple explicit metric types.

SpencerMalone:
Required three minor modifications to this test pattern:

- Couldn't keep using a map, as we needed duplicate entries in the data structure
- Added support for an explicit `metricType`, for when you care about that.
- Enabled the mapping cache to better simulate "real world" scenarios

Here's my test results from pre and post #229:

```
$ go test
WARN[0000] backtracking required because of match "backtrack.*.bbb", matching performance may be degraded  source="fsm.go:313"
WARN[0000] backtracking required because of match "*.dummy.*", matching performance may be degraded  source="fsm.go:313"
WARN[0000] match "order.*.*" is a super set of match "order.*.bbb" but in a lower order, the first will never be matched  source="fsm.go:294"
WARN[0000] match "test.foo.*" is a super set of match "test.foo.*" but in a lower order, the first will never be matched  source="fsm.go:294"
--- FAIL: TestMetricMapperYAML (0.00s)
    mapper_test.go:758: 21.'\x01': Expected name test_foo_sum, got test_foo_current
FAIL
exit status 1
FAIL	github.com/prometheus/statsd_exporter/pkg/mapper	0.016s

$ git status
HEAD detached at 50d5932
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   mapper_test.go

no changes added to commit (use "git add" and/or "git commit -a")

$ git pull origin master
From https://github.com/prometheus/statsd_exporter
 * branch            master     -> FETCH_HEAD
Updating 50d5932..5832aa9
<snip for length>

$ go test
WARN[0000] backtracking required because of match "backtrack.*.bbb", matching performance may be degraded  source="fsm.go:313"
WARN[0000] backtracking required because of match "*.dummy.*", matching performance may be degraded  source="fsm.go:313"
WARN[0000] match "order.*.*" is a super set of match "order.*.bbb" but in a lower order, the first will never be matched  source="fsm.go:294"
WARN[0000] match "test.foo.*" is a super set of match "test.foo.*" but in a lower order, the first will never be matched  source="fsm.go:294"
PASS
ok  	github.com/prometheus/statsd_exporter/pkg/mapper	0.016s

### DIFF
--- a/pkg/mapper/mapper_test.go
+++ b/pkg/mapper/mapper_test.go
@@ -18,12 +18,14 @@ import (
 	"time"
 )
 
-type mappings map[string]struct {
-	name       string
-	labels     map[string]string
-	quantiles  []metricObjective
-	notPresent bool
-	ttl        time.Duration
+type mappings []struct {
+	statsdMetric string
+	name         string
+	labels       map[string]string
+	quantiles    []metricObjective
+	notPresent   bool
+	ttl          time.Duration
+	metricType   MetricType
 }
 
 func TestMetricMapperYAML(t *testing.T) {
@@ -85,8 +87,9 @@ mappings:
 
   `,
 			mappings: mappings{
-				"test.dispatcher.FooProcessor.send.succeeded": {
-					name: "dispatch_events",
+				{
+					statsdMetric: "test.dispatcher.FooProcessor.send.succeeded",
+					name:         "dispatch_events",
 					labels: map[string]string{
 						"processor": "FooProcessor",
 						"action":    "send",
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: pkg/mapper/mapper_test.go [hunk 1]
```diff
@@ -491,6 +517,30 @@ mappings:
     `,
 			configBad: true,
 		},
+		// Config with multiple explicit metric types
```
SpencerMalone: Where my new test actually starts. Just adding a comment to increase visibility.

matthiasr:
thanks a lot!

PR #233, opened 2019-06-11
Use promhttp instead of http

### LINKED ISSUE #80: some common metrics names already in use by exporter
I'm not sure that I have good proposal to solve this, but I /just/ ran into the issue when mapping some statsd metrics that I cannot define a metric `http_requests_total` because it is already defined by default.

I suppose if an existing metric is redefined by the mapping that the user defined mapping should win out. I'm less sure what to do with the internal telemetry/instrumentation which had originally defined the metric.

cdambo:
client_golang's `http` is [deprecated](https://github.com/prometheus/client_golang/blob/8d9c21f0cf5f73718ee86bb3a8e6432b88e9291d/prometheus/http.go#L52) in favor of `promhttp`.
 
This should fix #80 

Tagging @matthiasr

### DIFF
--- a/main.go
+++ b/main.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/howeyc/fsnotify"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/prometheus/common/log"
 	"github.com/prometheus/common/version"
 	"gopkg.in/alecthomas/kingpin.v2"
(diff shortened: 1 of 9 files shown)


beorn7:
Looks good in general. I think you need to update the vendor dir to make the test pass.

beorn7:
Which you just tried.

But I think you didn't use `GO111MODULE=on go mod vendor` to do so.

beorn7:
In other words: You vendored more files than `go mod vendor` would do. And that's why the CI complains.

beorn7:
\o/ Test passed! @matthiasr I think this is good to go. (But it isn't really related to #80. That would require two different custom handlers, each handling a different registry: one the registry for statsd metrics, one the registry for exporter metrics.)

cdambo:
Thanks for the help @beorn7! 😄 

Right about #80 although the new `promhttp` doesn't use common metric names anymore, so one is able to initialize their own `http_requests_total`, for instance.

beorn7:
Got it. Yes, that's indeed true. The HTTP metrics of that kind had been deprecated.

But of course, you still get the fundamentally same problem if a statsd metric happens to match one of the standard go_... or process_... metrics or one of the metrics created by the exporter itself.

matthiasr:
Thank you! I think this solves the core of #80 since `http_requests_total` was the main source of conflicts as far as I know.

cdambo:
@matthiasr Hi 😄 
When do you plan to make the next release?

matthiasr:
it's done now!

PR #244, opened 2019-07-05
Quote all matchers in YAML examples

matthiasr:
They can start with a * which makes YAML barf if not quoted. Instead of
figuring out individually what does and does not need to be quoted,
always quote.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,14 +123,14 @@ An example mapping configuration:
 
 ```yaml
 mappings:
-- match: test.dispatcher.*.*.*
+- match: "test.dispatcher.*.*.*"
   name: "dispatcher_events_total"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -123,14 +123,14 @@ An example mapping configuration:
...
-- match: test.dispatcher.*.*.*
+- match: "test.dispatcher.*.*.*"
   name: "dispatcher_events_total"
   labels:
     processor: "$1"
     action: "$2"
     outcome: "$3"
     job: "test_dispatcher"
-- match: *.signup.*.*
+- match: "*.signup.*.*"
```
matthiasr: this one in particular was invalid.

PR #246, opened 2019-07-08
Fix TTL bug

cdnhaese:
Fixes a bug where setting the TTL parameter would periodically remove time series.
The reason for this was that the 'lastRegisteredAt' element was only set at creation of the time series, and not updated every time a new event was received for that time series.
__Maintainer:__ (@matthiasr)


matthiasr:
Thank you!

PR #261, opened 2019-09-03
Update use_ms note for dogstatsd client

### LINKED ISSUE #260: Confusing documentation regarding `use_ms`?
Hi all, the README has this bit:

> The DogStatsD client by default emits timer metrics in seconds, set use_ms to True to fix this.

Is this correct?

Looking at the dogstatsd Python client, we have this code in [base.py](https://github.com/DataDog/datadogpy/blob/master/datadog/dogstatsd/base.py):

```
def timing(self, metric, value, tags=None, sample_rate=1):
        """
        Record a timing, optionally setting tags and a sample rate.
        >>> statsd.timing("query.response.time", 1234)
        """
        self._report(metric, 'ms', value, tags, sample_rate)
```
This is reporting the metrics in `ms`.

amitsaha:
Fixes #260

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,7 +235,12 @@ mappings:
 
 Note that timers will be accepted with the `ms`, `h`, and `d` statsd types.  The first two are timers and histograms and the `d` type is for DataDog's "distribution" type.  The distribution type is treated identically to timers and histograms.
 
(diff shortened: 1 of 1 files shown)


matthiasr:
Wonderful, thank you!

PR #262, opened 2019-09-05
README - Add a note regarding unit conversions for timers

amitsaha:
While rolling out `statsd exporter`, this is one of the things that confused me. What's the final unit of the exported metric when I am querying prometheus? I took a look at the code to clarify, so thought adding this in the documentation is a good idea. 

Please take a look.


matthiasr:
Thank you!

Issue #250, opened 2019-07-22
Handle sampling factors on histograms


matthiasr:
The example given in the [dogstatsd docs](https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/) for a sampled histogram:

```
## Sample the song length histogram half of the time
song.length:240|h|@0.5
```

is accepted by the exporter, but the sampling is ignored:

```
DEBU[0008] Illegal sampling factor for non-counter metric on line song.length:240|h|@0.5  source="exporter.go:393"
```

We should allow sampling for non-counter metrics like this too; only gauges cannot be sampled.

amitsaha:
Just looking at this out of curiosity. Can you please explain the sampling logic here?
```
..
} else if statType == "ms" {
    multiplyEvents = int(1 / samplingFactor)
}
```
So, for a `samplingFactor` for a timing metric of of 0.5, `multiplyEvents` is 2. Then:

```
for i := 0; i < multiplyEvents; i++ {
			event, err := buildEvent(statType, metric, value, relative, labels)
			if err != nil {
				log.Debugf("Error building event on line %s: %s", line, err)
				sampleErrors.WithLabelValues("illegal_event").Inc()
				continue
			}
			events = append(events, event)
		}
```
Thus, we append the same event twice in the `events` array. Is it essentially "compensating" this metric by considering that the metric's value was reported to be the same value twice since it was already sampled by `statsd` and hence only one of the two events was considered?

matthiasr:
That's exactly right. We approximate the "real" metrics by generating
copies of the sampled event until it is (approximately) back to 1:1

amitsaha:
hi @matthiasr thanks, made an initial PR. wanted to understand your reasoning about not supporting `gauge`? Is it because of the inaccuracy that the "compensation" process might introduce?

matthiasr:
The value of a gauge is whatever value was set last, so I am not sure what
sampling would do here … we could consider *ignoring* it altogether though.

PR #264, opened 2019-09-09
Support sampling factor for all statsd metric types

amitsaha:
This is an attempted fix for #250

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -389,11 +389,7 @@ samples:
 			for _, component := range components[2:] {
 				switch component[0] {
 				case '@':
-					if statType == "g" {
-						log.Debugln("Illegal sampling factor for gauge metric type on line", line)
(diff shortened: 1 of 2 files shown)


exporter.go
```diff
@@ -389,8 +389,8 @@ samples:
 			for _, component := range components[2:] {
 				switch component[0] {
 				case '@':
-					if statType != "c" && statType != "ms" {
-						log.Debugln("Illegal sampling factor for non-counter metric on line", line)
+					if statType == "g" {
+						log.Debugln("Illegal sampling factor for gauge metric type on line", line)
```
matthiasr: I'm thinking about the semantics of a sampled gauge … I suppose it would make sense to set it to whatever the last submitted value is, no matter the sampling factor?

matthiasr:
Wonderful, thank you!

PR #267, opened 2019-09-14
Add librato tag support

twooster:
## What

Add support for Librato-style tags. These appear in the metric name rather than as a separate field, ala:

```
metric.name#tag=val,tag2=val2:0|c
```

## Why

Maybe this is just scratching my itch, but I'm facing the impossibility of attempting to incrementally move from Librato (and librato tags) to an internal Prometheus server. The only way forward for me is to either heavily hack the original Statsd server and the Librato backend plugin to support DogStatsD tags, or to add a bit of support to this exporter, which is what I've opted to do.

## Discussion

This is a fairly straightforward implementation, and is almost entirely backwards compatible, so I didn't bother to invoke the mailing list. The only place this might break is if someone is already using metrics with hash signs in them -- but this should probably not be occurring in most cases. If desired, I can add a configuration flag for this to prevent conflicts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,15 +29,37 @@ We recommend this only as an intermediate solution and recommend switching to
 [native Prometheus instrumentation](http://prometheus.io/docs/instrumenting/clientlibs/)
 in the long term.
 
-### DogStatsD extensions
+### Tagging Extensions
 
-The exporter will convert DogStatsD-style tags to prometheus labels. See
-[Tags](https://docs.datadoghq.com/developers/dogstatsd/data_types/#tagging) in the DogStatsD
-documentation for the concept description and
-[Datagram Format](https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/)
-for specifics. It boils down to appending
-`|#tag:value,another_tag:another_value` to the normal StatsD format.  Tags
-without values (`#some_tag`) are not supported.
+The exporter supports both Librato-style tags and DogStatsD-style tags,
+which will be converted into Prometheus labels.
+
+For Librato-style tags, they must be appended to the metric name, as so:
+
+```
+metric.name#tagName=val,tag2Name=val2:0|c
+```
+
+See the [statsd-librato-backend README](https://github.com/librato/statsd-librato-backend#tags)
+for a more complete description.
+
+For DogStatsD-style tags, they're appended as another section at the end of the
+metric, as so:
+
+```
+metric.name:0|c|#tagName=val,tag2Name=val2
+```
+
+See [Tags](https://docs.datadoghq.com/developers/dogstatsd/data_types/#tagging)
+in the DogStatsD documentation for the concept description and
+[Datagram Format](https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/).
+
+Although you can use both tag types simultaneously, this is not recommended.
+DogStatsD `name=value` pairs will take priority over Librato tags with the same
+name.
+
+For both Librato and DogStatsD tags, tags without values (`#some_tag`) are not
+supported.
 
 ## Building and Running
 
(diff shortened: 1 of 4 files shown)


matthiasr:
Thank you! This looks reasonable at first glance (TIL that there is another tagging style). I'll go over the code in detail soon.

twooster:
There's unfortunately many different tagging styles. :grinning: There's some discussion here:

https://github.com/statsd/statsd/issues/619

I could relatively simply also support [InfluxDB tagging style](https://www.influxdata.com/blog/getting-started-with-sending-statsd-metrics-to-telegraf-influxdb/#introducing-influx-statsd) (search for either '#' or ',' to begin first tagging), since that's probably also a reasonable and large target to hit? Let me know, and I'll update the PR.

matthiasr:
Fine with me if it's a straightforward change to you!

matthiasr:
OK and the SignalFX way is … [interesting](https://docs.signalfx.com/en/latest/integrations/agent/monitors/collectd-statsd.html#adding-dimensions-to-statsd-metrics) 😂 

I wonder at what point we outgrow if/else and should graduate to some kind of more formalized _parsing_?

twooster:
I've gone ahead and added support for InfluxDB style tags, and updated the documentation and tests to match. :ship:

README.md
```diff
@@ -29,15 +29,52 @@ We recommend this only as an intermediate solution and recommend switching to
...
+See [Tags](https://docs.datadoghq.com/developers/dogstatsd/data_types/#tagging)
+in the DogStatsD documentation for the concept description and
+[Datagram Format](https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/).
+Note that this tagging style is incompatible with the original `statsd`
+implementation, you will be unable to use it as a repeater in front of the
+statsd-to-prometheus exporter.
+
+Although you can use both name-appended tags (Librato or InfluxDB) and
+metric-appended tags simultaneously, this is not recommended. If you do, be
+aware that DogStatsD `name=value` pairs will take priority over Librato tags
```
matthiasr: I suppose they would also take priority over InfluxDB tags?

twooster: Sorry, yup, forgot to change that bit of documentation.

bridge_test.go
```diff
@@ -127,6 +127,46 @@ func TestHandlePacket(t *testing.T) {
...
+			out: Events{
+				&CounterEvent{
+					metricName: "foo",
+					value:      100,
+					labels:     map[string]string{"_09digits": "0", "tag_with_dots": "1"},
+				},
+			},
+		}, {
+			name: "influxdb tag extension",
+			in:   "foo#tag1=bar,tag2=baz:100|c",
```
matthiasr: this is repeating the librato test, did you mean

```suggestion
			in:   "foo,tag1=bar,tag2=baz:100|c",
```

bridge_test.go
```diff
@@ -127,6 +127,46 @@ func TestHandlePacket(t *testing.T) {
...
+			out: Events{
+				&CounterEvent{
+					metricName: "foo",
+					value:      100,
+					labels:     map[string]string{"tag1": "bar", "tag2": "baz"},
+				},
+			},
+		}, {
+			name: "influxdb tag extension with tag keys unsupported by prometheus",
+			in:   "foo#09digits=0,tag.with.dots=1:100|c",
```
matthiasr: ditto

```suggestion
			in:   "foo,09digits=0,tag.with.dots=1:100|c",
```

exporter.go
```diff
@@ -343,10 +371,22 @@ func lineToEvents(line string) Events {
...
+	metric := parseNameAndTags(elements[0], labels)
+
 	var samples []string
 	if strings.Contains(elements[1], "|#") {
-		// using datadog extensions, disable multi-metrics
+		// using DogStatsD tags
+
+		// don't allow mixed tagging styles
+		if len(labels) > 0 {
+			sampleErrors.WithLabelValues("malformed_line").Inc()
```
matthiasr: Let's count it separate from a generally malformed line:

```suggestion
			sampleErrors.WithLabelValues("mixed_tagging_styles").Inc()
```

bridge_test.go
twooster: Oops. Yes.

exporter.go
twooster: Works for me.

matthiasr:
Thank you! This is a great addition, and very nicely done.

PR #271, opened 2019-10-09
Add ability to expose server on unix socket

nabokihms:
@matthiasr 
Useful for installations where you hide metrics servers behind the proxy and don't want to expose extra ports.


brian-brazil:
Thanks for your PR, we explicitly only expose over TCP as that's how HTTP works.

nabokihms:
HTTP requires bidirectional connection, for example socket of type SOCK_STREAM, to work. It doesn't matter tcp socket or unix socket.

Let me describe my problem:
I have Kubernetes cluster and one pod with nginx and statsd exporter.
When I had enabled `hostNetwork: true`, statsd exporter port exposed to node IP.
So, I wish to mount emptyDir with unix socket to statsd and nginx containers to proxy_pass one of nginx locations to statsd exporter. This method allow us to expose only nginx port.

Maybe you can suggest another solution.

matthiasr:
I would rather not carry yet another way to bind to ports. We already have too many flags for all the statsd listen ports.

You can make the exporter listen on localhost, that way nginx can reach it but it is not reachable directly.

PR #272, opened 2019-10-10
Switch to go-kit logging

### LINKED ISSUE #270: Switch logging to go-kit
About two years ago we decided to move from logrus to go-kit for logging, this is one of the repos that still needs switching. prometheus/snmp_exporter#447 is an example PR.

mhartenbower:
Fixes #270


exporter.go
```diff
@@ -575,7 +581,7 @@ func (l *StatsDUnixgramListener) Listen() {
 			if strings.HasSuffix(err.Error(), "use of closed network connection") {
 				return
 			}
-			log.Fatal(err)
+			panic(err)
```
matthiasr: Please don't panic – log an error and then exit(1).

mhartenbower:
@matthiasr I'm not actually sure what the problem is at this point.

The build looks good. It's failing on `make common-unused`, but I'm having trouble following the output and I can't reproduce the issue locally. 

Mind taking a look?

matthiasr:
It looks like it thinks that the vendor directory and go modules are out of sync. Could you

- use the latest released Go
- rebase against mainline
- `go mod vendor`
- `go mod tidy`

and commit the changes that the last two commands make? You can also try running `make common-unused` locally, and committing the result.

brian-brazil:
Did you have a chance to look at the vendoring issues?

mhartenbower:
@brian-brazil Will get it resolved this weekend.

brian-brazil:
Did you get anywhere with this?

PR #275, opened 2019-11-05
Fix format for dogstatsd tags

ergo:
Only `key:value` format for dogstatsd is accepted by latest version of statsd exporter.
I've tried both official client and bash scripts to send the data:

    echo "newpure.test:1|c|@0.5|#country:china" | nc -w 1 -u 127.0.0.1 9125` - will work
    echo "newpure.test:1|c|@0.5|#country=china" | nc -w 1 -u 127.0.0.1 9125 - will not work

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ For DogStatsD-style tags, they're appended as a `|#` delimited section at the
 end of the metric, as so:
 
 ```
-metric.name:0|c|#tagName=val,tag2Name=val2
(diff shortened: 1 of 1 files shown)


matthiasr:
Ah yes, thank you for spotting this! Can you please add the DCO sign-off as described in the failing test?

ergo:
Sure, sorry I missed that, the patch is now signed.

matthiasr:
Thank you!

PR #277, opened 2019-11-26
Updated docs and added good/bad example yaml

### LINKED ISSUE #274: Regex example not working
Hello, I have tried to run a docker container using as statsd mapping configuration file the one that you propose for the regex example:

```yaml
mappings:
- match: "test\.(\w+)\.(\w+)\.counter"
  match_type: regex
  name: "${2}_total"
  labels:
    provider: "$1"
```
and it fails, logging the following error:
```level=fatal msg="Error loading config:yaml: line 2: found unknown escape character" source="main.go:245"```

I thought it was a GO issue because of the "\\" characters used in the regex chain, so I scaped them using double "\\". However, although the container starts, the regex does not work, because it does not catch what it is supposed to catch. 

Could you please help me with this issue? Thanks in advance.

### PR DESCRIPTION
**Pull request by xocasdashdash**
This should fix #274 and also some tests with good/bad yaml to verify this behaviour.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,5 @@
 ## unreleased / ????-??-??
 
-* [ENHANCEMENT] Updated readme docs and added a test for invalid yaml ([#274](https://github.com/prometheus/statsd_exporter/issues/274))
 * [ENHANCEMENT] Support sampling factors for all statsd metric types ([#264](https://github.com/prometheus/statsd_exporter/issues/250))
 * [ENHANCEMENT] Support Librato and InfluxDB labeling formats ([#267](https://github.com/prometheus/statsd_exporter/pull/267))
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: CHANGELOG.md [hunk 1]
```diff
@@ -1,5 +1,6 @@
 ## unreleased / ????-??-??
 
+* [ENHANCEMENT] Updated readme docs and added a test for invalid yaml ([#274](https://github.com/prometheus/statsd_exporter/issues/274))
```
matthiasr: I usually don't include documentation and test changes in the changelog, because I want to focus on what is important for someone who is already using it in production. Could you remove this please?

matthiasr: (the README is not released as part of the tarball, so it doesn't really have _versioning_ per se)

xocasdashdash: Sure!

matthiasr:
Thank you!

PR #280, opened 2020-01-02
mapper cache: Add cache metrics for total gets and hits

bakins:
Add metrics to record mapper cache hits and total gets.

### DIFF
--- a/pkg/mapper/mapper_cache.go
+++ b/pkg/mapper/mapper_cache.go
@@ -25,6 +25,18 @@ var (
 			Help: "The count of unique metrics currently cached.",
 		},
 	)
+	cacheGetsTotal = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Name: "mapping_cache_gets_total",
+			Help: "The count of total cache gets.",
+		},
+	)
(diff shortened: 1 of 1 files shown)


brian-brazil:
Separate hits and total metrics are easier to work with, as you don't have to add before you divide.

bakins:
@brian-brazil should we also have a separate metric for total cache Get requests?

brian-brazil:
The total is the get requests.

bakins:
> The total is the get requests.

Not sure I follow. Do we already have a metric that tracks total cache gets?

bakins:
Refactored to have a metric for total gets and a metric for total hits.

Sorry, my mind is apparently still on holiday as it took me a bit to realize what you were suggesting @brian-brazil

brian-brazil:
:+1:

bakins:
@matthiasr changed the metrics names.

Note: the existing metric in mapper is named `statsd_exporter_cache_length` https://github.com/prometheus/statsd_exporter/blob/2e3d3ab962ec7c305e590dec4ffebc465a00c034/pkg/mapper/mapper_cache.go#L24

PR #282, opened 2020-01-10
Rename mapping cache length metric

matthiasr:
The mapper package can be used outside the statsd exporter, so it is
better to use generic names here.

follow-up to #280

### DIFF
--- a/pkg/mapper/mapper_cache.go
+++ b/pkg/mapper/mapper_cache.go
@@ -21,7 +21,7 @@ import (
 var (
 	cacheLength = prometheus.NewGauge(
 		prometheus.GaugeOpts{
-			Name: "statsd_exporter_cache_length",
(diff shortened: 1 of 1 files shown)


brian-brazil:
> The mapper package can be used outside the statsd exporter, so it is better to use generic names here.

Whether it can be used elsewhere is irrelevant to metric naming. For example even if you use the Prometheus tsdb elsewhere, the correct metric name prefix is still `prometheus_tsdb_`

matthiasr:
then let me rephrase that. this package is a generic mapper from one metric model into another. it is not _exporter_ specific.

matthiasr:
therefore, all metrics are now under the `mapping_cache` prefix, wherever they are used.

brian-brazil:
That's a very generic name, there should probably be a statsd or prometheus prefix in there somewhere.

Issue #263, opened 2019-09-07
Offline configuration check


McLeod095:
Hello, please add check config for daemon with check map file.
For manual check before restart prod daemon.

matthiasr:
That's a good idea! I think this doesn't have to be very complicated, I wouldn't add a second binary just for this; IMO a flag that causes it to load the config and then exit would be enough.

PR #283, opened 2020-01-10
PR #272 continued

matthiasr:
Picking up #272 to get it over the finish line:

* [x] fix the build issues
* [x] convert panics to log+exit (missed this in initial review)

cc @mhartenbower


matthiasr:
note for the changelog: the logger change is only mildly breaking – the valid values for `--logger.format` have changed (`[logfmt, json]`) and the `fatal` log level is no longer valid.

main.go
```diff
@@ -248,12 +270,16 @@ func main() {
...
+			os.Exit(1)
 		}
 		if *dumpFSMPath != "" {
 			err := dumpFSM(mapper, *dumpFSMPath, logger)
 			if err != nil {
-				panic(fmt.Sprintf("Error dumping FSM: %s", err))
+				level.Error(logger).Log("msg", "error dumping FSM", "error", err)
+				// Failure to dump the FSM is an error (the user asked for it and it
+				// didn't happen) but not fatal (the exporter is fully functional
+				// afterwards).
```
brian-brazil: As an aside, should this always exit anyway? This sounds more like a tool than what a daemon should do in passing.

brian-brazil:
:+1:

main.go
matthiasr: hmm it could, yeah, although then I'd prefer to start a mapper CLI which could also do config validation.

matthiasr: cf. #263

PR #284, opened 2020-01-10
Make mapper cache metric names more specific

matthiasr:
cf. https://github.com/prometheus/statsd_exporter/pull/282#issuecomment-573031514


brian-brazil:
:+1:

Issue #273, opened 2019-10-14
Metric mapping cache issue when variable uses in name


nikitka:
Hello,

I've found a issue with mapping cache when variable uses in name. Trivial example:

```
mappings:
  - match: aa.bb.*.*
    name: "aa_bb_${1}_total"
    labels:
      app: "$2"
```

```
echo "aa.bb.cc.myapp:1|c\naa.bb.cc.myapp:1|c\naa.bb.dd.myapp:1|c\naa.bb.cc.myapp:1|c\n" | nc -w 1 -u 127.0.0.1 9125 && curl -s http://127.0.0.1:9102/metrics|grep myapp
aa_bb_cc_total{app="myapp"} 2
aa_bb_dd_total{app="myapp"} 2
```

When cache disabled (via `statsd.cache-size=0`) I got expected results:

```
echo "aa.bb.cc.myapp:1|c\naa.bb.cc.myapp:1|c\naa.bb.dd.myapp:1|c\naa.bb.cc.myapp:1|c\n" | nc -w 1 -u 127.0.0.1 9125 && curl -s http://127.0.0.1:9102/metrics|grep myapp
aa_bb_cc_total{app="myapp"} 3
aa_bb_dd_total{app="myapp"} 1
```

If I understand correctly `FSM.GetMapping` returns the same mappingState pointer for `aa.bb.cc.myapp` and `aa.bb.dd.myapp`.  And before placing in the cache we update `Name` property for both metrics (because pointer is the same). I see solution with copying mappingState before placing in cache, but not sure of the correctness and I don't have Go language skills to make it perfectly. Could you fix this please?

matthiasr:
Thank you for the report and the complete repro instructions!

matthiasr:
@bakins is this something you could take a look at?

PR #286, opened 2020-01-11
Use correct name when multiple names match same FSM match

bakins:
Fixes https://github.com/prometheus/statsd_exporter/issues/273

Make shallow copy of FSM result so that we do not overwrite name.

Before shallow copy:

```
$ go test -v -run='^(TestMultipleMatches)$' ./pkg/mapper/
=== RUN   TestMultipleMatches
--- FAIL: TestMultipleMatches (0.00s)
    mapper_test.go:936: 1:1 Expected name aa_bb_dd_total, got aa_bb_cc_total
FAIL
FAIL	github.com/prometheus/statsd_exporter/pkg/mapper	0.260s
FAIL
```

After shallow copy:
```
$ go test -v -run='^(TestMultipleMatches)$' ./pkg/mapper/
=== RUN   TestMultipleMatches
--- PASS: TestMultipleMatches (0.00s)
PASS
ok  	github.com/prometheus/statsd_exporter/pkg/mapper	(cached)
```

### DIFF
--- a/pkg/mapper/mapper.go
+++ b/pkg/mapper/mapper.go
@@ -245,7 +245,8 @@ func (m *MetricMapper) GetMapping(statsdMetric string, statsdMetricType MetricTy
 	if m.doFSM {
 		finalState, captures := m.FSM.GetMapping(statsdMetric, string(statsdMetricType))
 		if finalState != nil && finalState.Result != nil {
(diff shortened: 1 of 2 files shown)


matthiasr:
Thank you!

Issue #234, opened 2019-06-11
Split into reusable packages


Gollam:
There are some useful structs and functions in the exporter.go file, which I'd like to reuse in my project. But I cannot import the file this way, because it is part of the main package. Can you please move it to some other package?

matthiasr:
Yes, I think this is a really good idea, I have been thinking about this as well.

I am also maintaining the [graphite exporter](https://github.com/prometheus/graphite_exporter). The two do almost _exactly_ the same, the only difference is the parsing of the metric lines – alas, the code has diverged so much that they are almost completely different. The statsd exporter is much more advanced and has had many performance improvements, so it could be the basis for both.

Specifically, the bits that I would like to see extracted are:

* listening on ports and handling the stream(s) of line-wise input
* statsd metric line parsing from `[]byte` to a statsd-semantics event (dot-style metric-name + label/value pairs + type)
* the mapping from statsd-semantics events to prometheus metrics is already extracted and reused
* the registry, with expiration + metrics exposition

With this, the main package should only consist of plugging these together, and then the graphite exporter could do the same except plug in a different line parser.

What is your use case? Which bits would you need exported?

Gollam:
Mainly, what I need is the escapeMetricName and lineToEvents functions, and of course the related structs. With those, I could skip writing a parser for the protocol.

matthiasr:
Hmm, I am wondering if escapeMetricName should be part of the mapping logic, since it's essentially the fallback mapping mechanism.

matthiasr:
Done in #298

PR #289, opened 2020-01-16
Move escapeMetricName to mapper

bakins:
https://github.com/prometheus/statsd_exporter/issues/234#issuecomment-501206903

I've copy pasted this into multiple personal projects as well, so let's export this.

We could integrate it more into mapping later.

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -54,61 +54,6 @@ type Exporter struct {
 	logger   log.Logger
 }
 
-// Replace invalid characters in the metric name with "_"
-// Valid characters are a-z, A-Z, 0-9, and _
-func escapeMetricName(metricName string) string {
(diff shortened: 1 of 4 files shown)


exporter.go
```diff
@@ -159,14 +104,14 @@ func (b *Exporter) handleEvent(event Event) {
 			errorEventStats.WithLabelValues("empty_metric_name").Inc()
 			return
 		}
-		metricName = escapeMetricName(mapping.Name)
```
matthiasr: is there anything that prevents us from making the mapper return an escaped name in the first place?

matthiasr:
the new files are lacking the license header, could you add it please?

matthiasr:
Thank you!

PR #291, opened 2020-01-28
[DRAFT] move telemetry and expiring registry to separate packages

glightfoot:
This moves the expiring registry to a separate package so that it can be reused in other projects where an expiring registry might be useful. In addition, it moves telemetry out of the main package and into telemetry since it was low hanging fruit (I can remove this part if it's too much for one PR).

I hacked this up pretty quickly from an internal adaptation of the registry that only used counters and gauges and removed a lot of the defaults, so I am not sure if I got the summary and histogram logic correct when changing from passing in the mapping to a more general API, so double-checking would be greatly appreciated! Additionally, I added a mutex for the expiring registry so it can be used concurrently. The internal adaptation it is based on seems to run fine and the race detector doesn't complain, but I'm not super familiar with the code so I might have missed a lock somewhere. As used in the statsd_exporter, the lock is not necessary, which might add a slight performance hit.

Just looking for some feedback on the general approach and the API for the expiring registry. If all looks good, I'll work on improving the documentation and adding more tests for the new package.

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -27,9 +27,10 @@ import (
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/prometheus/client_golang/prometheus"
-
 	"github.com/prometheus/statsd_exporter/pkg/clock"
+	"github.com/prometheus/statsd_exporter/pkg/expiringregistry"
 	"github.com/prometheus/statsd_exporter/pkg/mapper"
+	"github.com/prometheus/statsd_exporter/pkg/telemetry"
 )
 
 const (
@@ -50,7 +51,7 @@ func (u uncheckedCollector) Collect(c chan<- prometheus.Metric) {
 
 type Exporter struct {
 	mapper   *mapper.MetricMapper
-	registry *registry
+	registry *expiringregistry.Registry
 	logger   log.Logger
 }
 
@@ -62,7 +63,7 @@ func (b *Exporter) Listen(e <-chan Events) {
 	for {
 		select {
 		case <-removeStaleMetricsTicker.C:
-			b.registry.removeStaleMetrics()
+			b.registry.RemoveStaleMetrics()
 		case events, ok := <-e:
 			if !ok {
 				level.Debug(b.logger).Log("msg", "Channel is closed. Break out of Exporter.Listener.")
@@ -87,7 +88,7 @@ func (b *Exporter) handleEvent(event Event) {
 	}
 
 	if mapping.Action == mapper.ActionTypeDrop {
-		eventsActions.WithLabelValues("drop").Inc()
+		telemetry.EventsActions.WithLabelValues("drop").Inc()
 		return
 	}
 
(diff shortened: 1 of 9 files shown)


brian-brazil:
> in other projects where an expiring registry might be useful.

Can you expand on where you see this might be useful?

glightfoot:
Sure! In this case, it was an internal exporter checking on external resources that can appear and disappear over longer timeframes. Without the expiring registry, stale metrics were served until the exporter was restarted. I chose to adapt this instead of reinventing the wheel since it did what I needed.

matthiasr:
I also want to use this in the graphite exporter. it has the same needs but
a completely separately evolved implementation

exporter.go
```diff
@@ -62,7 +63,7 @@ func (b *Exporter) Listen(e <-chan Events) {
 	for {
 		select {
 		case <-removeStaleMetricsTicker.C:
-			b.registry.removeStaleMetrics()
+			b.registry.RemoveStaleMetrics()
```
matthiasr: could this be part of a maintenance routine internal to the expiring registry? That way we would not have to rely on the library users to do it regularly.

exporter.go
```diff
@@ -159,23 +160,23 @@ func (b *Exporter) handleEvent(event Event) {
 
 		switch t {
 		case mapper.TimerTypeHistogram:
-			histogram, err := b.registry.getHistogram(metricName, prometheusLabels, help, mapping)
+			histogram, err := b.registry.GetHistogram(metricName, prometheusLabels, help, mapping.Buckets, mapping.Ttl)
```
matthiasr: conflict warning: this will change slightly with #290 since we are about to move the Buckets and Quantiles into a HistogramOptions/SummaryOptions sub-struct. It will be easy to adapt though.

pkg/expiringregistry/expiringregistry.go
```diff
@@ -276,10 +315,9 @@ func (r *registry) getHistogram(metricName string, labels prometheus.Labels, hel
 
 	var histogramVec *prometheus.HistogramVec
 	if vh == nil {
-		metricsCount.WithLabelValues("histogram").Inc()
-		buckets := r.mapper.Defaults.Buckets
-		if mapping.Buckets != nil && len(mapping.Buckets) > 0 {
-			buckets = mapping.Buckets
+		r.metricsCount.WithLabelValues("histogram").Inc()
+		if buckets == nil || len(buckets) == 0 {
+			buckets = r.defaults.Buckets
```
matthiasr: hmmmm so this is where the mapping and the expiration got very intertwined. What would the mapper need to return to make it unnecessary to handle defaults here? Would it make sense if the mapper returns a complete `prometheus.HistogramOpts` (and `prometheus.SummaryOpts` if timer type is summary) every time?

Conceptually, I think expiring registry needs to know about _Prometheus_ domain objects (metric names and types, labels and values, help, summary/histogram options) and _time_ (so that it can expire the former). It _should_ not need to care how they came to into being. Consequently, ideally the `expiringregistry` package should not have a dependency on the `mapper` package.

matthiasr:
Unfortunately we could not finish this PR in time, and it was overtaken by #298. The registry is now available at `pkg/registry` as a relatively direct extraction.

I am still interested in cleaning up the interfaces and responsibilities between the packages. I am going to close this PR since it is hopelessly merge-conflicted. If you (or anyone) has ideas how to improve the packages, please do file issues and PRs!

PR #298, opened 2020-03-12
:Issue 234: Split into reusable packages.

davidsonff:
Signed-off-by: Frank Davidson <frank_davidson@manulife.com>

Hi,

I tried to deal with Issue #234. I don't have access to a StatsD server or anything (I'm actually working towards fixing the similar Graphite Exporter problem).

This passed all the tests. Let me know if there are things to fix here.

Thank you for all your hard work!

Frank


SuperQ:
It looks like you're using an editor or environment that is re-formatting all files with DOS carriage-return. Please update your environment settings and revert those changes.

davidsonff:
Just tried to fix the line endings.

main.go
```diff
@@ -218,8 +330,8 @@ func main() {
 		}
 		defer tconn.Close()
 
-		tl := &StatsDTCPListener{conn: tconn, eventHandler: eventQueue, logger: logger}
-		go tl.Listen()
+		tl := &listener.StatsDTCPListener{Conn: tconn, EventHandler: eventQueue, Logger: logger}
+		go tl.Listen(linesReceived, eventsFlushed, tcpConnections, tcpErrors, tcpLineTooLong, *sampleErrors, samplesReceived, tagErrors, tagsReceived)
```
matthiasr: Stylistic question: is it a common pattern to pass the counters like this? It seems a bit prone to accidentally messing up the order and counting wrongly … would it make sense to make them fields in the StatsdDTCPListener struct?

pkg/event/event.go~
```diff
@@ -0,0 +1,134 @@
+// Copyright 2013 The Prometheus Authors
```
matthiasr: is seems like a few `*~` files have snuck in, would you mind removing them and adding a gitignore pattern (or possibly better, add this to your own global gitignore, since it's your environment that creates them?)

matthiasr:
Thank you so much for tackling this!

The ~ files are making the diff tricky, I'm going to do another review pass when those are out. Could you also take a look at the failing tests?

main.go
davidsonff: Probably! Let me know if everything else looks ok and I can try to get that in there... I'm sure there's a lot of sub-optimal stuff as I'm just getting into the code!

pkg/util/util.go
```diff
@@ -0,0 +1,66 @@
+// Copyright 2013 The Prometheus Authors
```
matthiasr: let's not start a "util" kitchen-sink package – all this does is parsing addresses so I would call it "address" or something. Content-wise this package is fine, but names have power and if we call it util it will accumulate all kinds of stuff because we stop forcing ourselves to think about proper package scopes in the future.

exporter_test.go
```diff
@@ -24,7 +24,12 @@ import (
 	dto "github.com/prometheus/client_model/go"
 
 	"github.com/prometheus/statsd_exporter/pkg/clock"
+	"github.com/prometheus/statsd_exporter/pkg/event"
+	"github.com/prometheus/statsd_exporter/pkg/exporter"
```
matthiasr: it seems strange to me that the exporter _tests_ are not inside the exporter package. Could you move those that make sense in too, and move the others to the appropriate place? it's fine to have program-wide tests in the main package but I would like them to be named appropriately then.

davidsonff:
Working on all of this. Might take a bit of time... Busy!

matthiasr:
awesome, thank you for checking in!

davidsonff:
I don't seem to be able to get the DCO check to work. Not sure what the issue is.

matthiasr:
I think it is taking issue with your _author_ line:

```
commit 72c05d2e06e555b3011739d7416ce1bf9d92b2a8 (HEAD -> issue-234)
Author: = <=>
Date:   Wed Apr 8 14:59:19 2020 -0400

    Updated structs and tests.

    Signed-off-by: Frank Davidson <davidfr@americas.manulife.net>
    Signed-off-by: Frank Davidson <ffdavidson@gmail.com>
```

something seems to be off with your gitconfig.

matthiasr:
Looks good! I want to merge this as is. We may iterate on the interfaces later; I'm not super happy with the amount of boilerplate around the metrics but also don't have a solution yet.

When fixing the DCO, could you also squash all the commits into one so we don't end up with the .exe in the history?

main.go
```diff
@@ -247,7 +386,7 @@ func main() {
 			}
 		}
 
-		ul := &StatsDUnixgramListener{conn: uxgconn, eventHandler: eventQueue, logger: logger}
+		ul := &listener.StatsDUnixgramListener{Conn: uxgconn, EventHandler: eventQueue, Logger: logger}
```
matthiasr, line 389: the unixgram listener is missing the full initialization with metrics, leading to a nil pointer dereference when the first packet is received there. could you add this please?

matthiasr:
Thank you so much for getting this done!

davidsonff:
@matthiasr I think maybe I should pass line.LineToEvents() as a function value into the listener package, i.e., pass the LineToEvent function as an argument into listener.HandleConn() for use by event.EventHandler()? I don't think I can replace it as is now in the Graphite Exporter? What do you think? That would make LineToEvents pretty swappable for different use cases?

matthiasr:
That's a good idea, but does the listener have to know about events at all? I think we can disentangle the listener, event parsing, and collector even further. The listener package could only deal in "packages" (single lines or bundles of lines?), sending them through a channel to a parser goroutine (possibly multiple).

The chain would roughly be listener -> (statsd / graphite / influxdb / ???) parser -> event queue -> collector.

We already have a well defined interface from the parser to the event queue. What would be the best way to hand unparsed input to the parser? byte slices or strings? Single lines only, or whole packets? what would "a whole packet" mean in the context of TCP though?

PR #306, opened 2020-05-14
Set Flush Interval on EventQueue

chai-nadig:
* Set Flush Interval on EventQueue

@matthiasr


matthiasr:
Thank you, and sorry for the delay!

PR #314, opened 2020-06-11
[307] Add support for histograms and distributions without unit conversion

### LINKED ISSUE #307: Support for non-timer histograms?
I'm working with a legacy application that emits StatsD metrics. Some of the metrics are histograms (i.e. 'h' type) and represent data like document page counts or request/response size. statsd-exporter assumes these are timers, and divides the values by 1000.

Is there any possibility this behavior can be made configurable?

glightfoot:
This fixes #307 by moving millisecond-->second conversion to the buildEvent function for `ms` statTypes and also adds a case for `h` and `d` statTypes without the conversion. This effectively makes the TimerEvent value in seconds/unitless. Since this is no longer just limited to timers, should the TimerEvent be renamed to DistributionEvent or something similar? 

Test cases were also added for the three distribution types and the conversions.

Summary:
`ms` units in milliseconds, converted to seconds for prometheus
`h` no units, no conversion
`d` no units, no conversion

The benchmark tests were broken, so the metrics were added back to them to get the benchmark working. That commit can be removed if necessary.
Before:
```
$ go test -bench=.
goos: darwin
goarch: amd64
BenchmarkUDPListener1-12                   79335             12802 ns/op
BenchmarkUDPListener5-12                    3866            304820 ns/op
BenchmarkUDPListener50-12                     33          32112409 ns/op
BenchmarkExporterListener-12                 183           6583010 ns/op
PASS
ok      github.com/prometheus/statsd_exporter   5.714s

```

After:
```
$ go test -bench=.
goos: darwin
goarch: amd64
BenchmarkUDPListener1-12                   80152             13512 ns/op
BenchmarkUDPListener5-12                    3638            331360 ns/op
BenchmarkUDPListener50-12                     31          34365441 ns/op
BenchmarkExporterListener-12                 183           6347688 ns/op
PASS
ok      github.com/prometheus/statsd_exporter   5.548s
```
 These results show no significant increase in execution time, as the difference is within the variation observed on repeated benchmarking.

### DIFF
--- a/bridge_test.go
+++ b/bridge_test.go
@@ -85,7 +85,7 @@ func TestHandlePacket(t *testing.T) {
 			out: event.Events{
 				&event.TimerEvent{
 					TMetricName: "foo",
-					TValue:      0.2, // convert statsd ms to seconds in parsing
+					TValue:      0.2,
 					TLabels:     map[string]string{},
 				},
 			},
@@ -448,7 +448,7 @@ func TestHandlePacket(t *testing.T) {
 			out: event.Events{
 				&event.TimerEvent{
 					TMetricName: "foo",
-					TValue:      200, // convert statsd ms to seconds in parsing
+					TValue:      200,
 					TLabels:     map[string]string{},
 				},
 			},
--- a/bridge_test.go
+++ b/bridge_test.go
@@ -83,60 +83,60 @@ func TestHandlePacket(t *testing.T) {
 			name: "simple timer",
 			in:   "foo:200|ms",
 			out: event.Events{
-				&event.TimerEvent{
-					TMetricName: "foo",
-					TValue:      0.2,
-					TLabels:     map[string]string{},
+				&event.ObserverEvent{
+					OMetricName: "foo",
+					OValue:      0.2,
+					OLabels:     map[string]string{},
 				},
 			},
 		}, {
 			name: "simple histogram",
 			in:   "foo:200|h",
 			out: event.Events{
-				&event.TimerEvent{
-					TMetricName: "foo",
-					TValue:      200,
-					TLabels:     map[string]string{},
+				&event.ObserverEvent{
+					OMetricName: "foo",
+					OValue:      200,
+					OLabels:     map[string]string{},
 				},
 			},
 		}, {
 			name: "simple distribution",
 			in:   "foo:200|d",
 			out: event.Events{
-				&event.TimerEvent{
-					TMetricName: "foo",
-					TValue:      200,
-					TLabels:     map[string]string{},
+				&event.ObserverEvent{
+					OMetricName: "foo",
+					OValue:      200,
+					OLabels:     map[string]string{},
 				},
 			},
 		}, {
 			name: "distribution with sampling",
 			in:   "foo:0.01|d|@0.2|#tag1:bar,#tag2:baz",
 			out: event.Events{
-				&event.TimerEvent{
-					TMetricName: "foo",
-					TValue:      0.01,
-					TLabels:     map[string]string{"tag1": "bar", "tag2": "baz"},
+				&event.ObserverEvent{
+					OMetricName: "foo",
+					OValue:      0.01,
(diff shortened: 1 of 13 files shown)


glightfoot:
@matthiasr

bridge_test.go
```diff
@@ -422,6 +422,36 @@ func TestHandlePacket(t *testing.T) {
...
+					TLabels:     map[string]string{},
+				},
+			},
+		}, {
+			name: "distribution with no unit conversion",
+			in:   "foo:200|d",
+			out: event.Events{
+				&event.TimerEvent{
+					TMetricName: "foo",
+					TValue:      200, // convert statsd ms to seconds in parsing
```
matthiasr, line 451: I think this comment is out of date now
```suggestion
					TValue:      200,
```

glightfoot: Yeah, sorry forgot to remove that when I was fixing benchmarks

glightfoot:
ObserverEvent makes sense and is nice and consistent. I pushed up that change now. The only thing I can think of now is that there are plenty of stats calls that still reference timer event:
```				
b.EventStats.WithLabelValues("timer").Inc()
```

I think it makes sense to change those to "observer", but that would be another breaking change to metrics usage, so let me know what you'd like to see.

glightfoot:
The other thing I can see is there are still lots of references to Timer in the Mapper pkg. I can change them, but then that breaks the configs since things like `timer_type` in the defaults would become `observer_type`, which would need some explaining in the documentation

matthiasr:
Ah, you ask all the good questions. I'm fine with "breaking" the metrics, a note in the changelog will help, which I will add anyway.

I am less willing to break configurations, how hard would it be to change the primary option but maintain compatibility? (i.e. document `observer_type`, deprecate `timer_type` but treat them the same)?

glightfoot:
Ok I'll change the metrics now. I haven't looked too much at the mapper config code, so I am not sure how simple of a change that will be

glightfoot:
I added some custom unmarshalers to handle both keys. `observer_type` will always override `timer_type`. Let me know what you think about this approach. Also note that while this passes all test right now, I haven't yet compiled it and run a full test

glightfoot:
I built this branch and tested with the following:

```
$ cat << EOF > test-config.yaml
defaults:
  timer_type: summary
mappings:
- match: "(.*)"
  name: "$1"
  match_type: regex
EOF

$ ./statsd_exporter --statsd.mapping-config=test-config.yaml &

$ for x in `seq 1 10`; do
    echo "foo_ms:200|ms" | nc -w 1 -cu localhost 9125
    echo "foo_h:200|h" | nc -w 1 -cu localhost 9125
    echo "foo_d:200|d" | nc -w 1 -cu localhost 9125
done

$ curl -s localhost:9102/metrics | grep -E "foo|statsd_exporter_events_total"
# HELP foo_d Metric autogenerated by statsd_exporter.
# TYPE foo_d summary
foo_d{quantile="0.5"} 200
foo_d{quantile="0.9"} 200
foo_d{quantile="0.99"} 200
foo_d_sum 2000
foo_d_count 10
# HELP foo_h Metric autogenerated by statsd_exporter.
# TYPE foo_h summary
foo_h{quantile="0.5"} 200
foo_h{quantile="0.9"} 200
foo_h{quantile="0.99"} 200
foo_h_sum 2000
foo_h_count 10
# HELP foo_ms Metric autogenerated by statsd_exporter.
# TYPE foo_ms summary
foo_ms{quantile="0.5"} 0.2
foo_ms{quantile="0.9"} 0.2
foo_ms{quantile="0.99"} 0.2
foo_ms_sum 1.9999999999999998
foo_ms_count 10
# HELP statsd_exporter_events_total The total number of StatsD events seen.
# TYPE statsd_exporter_events_total counter
statsd_exporter_events_total{type="observer"} 30
```

Is there anything else necessary for this PR?

matthiasr:
This is awesome, thank you!

PR #315, opened 2020-06-16
[268] Support SignalFx Tags

### LINKED ISSUE #268: Support SignalFX tagging style
As a follow-up to #267 we should also support the [SignalFX dimension statsd extension](https://docs.signalfx.com/en/latest/integrations/agent/monitors/collectd-statsd.html#adding-dimensions-to-statsd-metrics).

See also statsd/statsd#619 for more discussion around tagging styles.

This tagging style is somewhat different from the ones we already support. At this point it would be good to break "generalized statsd parsing" out into its own package, and make sure it is well tested (and benchmarked – this is a very hot code path).

glightfoot:
Adds parsing for SignalFX-style tagsand a benchmark for LineToEvents. Fixes #268. It's unclear from the signalfx docs, but it appears that the tags can exist anywhere in the name field, making parsing a little more complicated.

There is a definite slight cost to parsing these tags since they can be in the middle of the name, and have opening and closing runes.

Before:
```
go test -bench='BenchmarkLineToEvents(1|5|50)'                                                                                                                
goos: darwin
goarch: amd64
BenchmarkLineToEvents1-12         148065              7847 ns/op
BenchmarkLineToEvents5-12          30534             39436 ns/op
BenchmarkLineToEvents50-12          3038            392718 ns/op
PASS
ok      github.com/prometheus/statsd_exporter   4.261s

```

After:
```
go test -bench='BenchmarkLineToEvents(1|5|50)'                                                                                                                 
goos: darwin
goarch: amd64
BenchmarkLineToEvents1-12         132652              8131 ns/op
BenchmarkLineToEvents5-12          29119             40593 ns/op
BenchmarkLineToEvents50-12          2988            405923 ns/op
PASS
ok      github.com/prometheus/statsd_exporter   4.357s

```

### DIFF
--- a/bridge_test.go
+++ b/bridge_test.go
@@ -23,6 +23,7 @@ import (
 	"github.com/go-kit/kit/log"
 	"github.com/prometheus/client_golang/prometheus"
 	dto "github.com/prometheus/client_model/go"
+
 	"github.com/prometheus/statsd_exporter/pkg/clock"
 	"github.com/prometheus/statsd_exporter/pkg/event"
 	"github.com/prometheus/statsd_exporter/pkg/exporter"
@@ -282,6 +283,10 @@ func TestHandlePacket(t *testing.T) {
 		}, {
 			name: "librato/dogstatsd mixed tag styles without sampling",
 			in:   "foo#tag1=foo,tag3=bing:100|c|#tag1:bar,#tag2:baz",
+			out:  event.Events{},
+		}, {
+			name: "signalfx/dogstatsd mixed tag styles without sampling",
+			in:   "foo[tag1=foo,tag3=bing]:100|c|#tag1:bar,#tag2:baz",
 			out:  event.Events{},
 		}, {
 			name: "influxdb/dogstatsd mixed tag styles without sampling",
--- a/bridge_test.go
+++ b/bridge_test.go
@@ -221,6 +221,26 @@ func TestHandlePacket(t *testing.T) {
 				},
 			},
 		}, {
+			name: "SignalFx tag extension, missing closing bracket",
+			in:   "[tag1=bar,tag2=bazfoo.test:100|c",
+			out: event.Events{
+				&event.CounterEvent{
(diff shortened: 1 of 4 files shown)


glightfoot:
Note: if we move all the metrics into their own package, we can have the benchmark tests live in the package dirs instead at the top-level. Happy to do that too if necessary

glightfoot:
@matthiasr

glightfoot:
As-is, this does not check for any other tag styles if it find SignalFx tags

glightfoot:
Without knowing the acceptable characters, I thought the best way to handle it would be to pass the name on unchanged, but then the other tag parsing catches the ',' and splits it up trying to parse InfluxDB tags. If we want to short-circuit tag parsing if there's a non-matching bracket, I can do that, but do we want to drop the line, or just pass it through? In theory, you could have a valid InfluxDB tagged line with a bracket in the name... `foo[strange.name,09digits=0,tag.with.dots=1:100|c`

matthiasr:
I think the best would be passing it through, with a debug log line + metric increment if it _almost_ looks like signalfx tags but isn't. that way, existing behavior is unchanged for anything that isn't a fully formed signalfx tag.

I would rather not have the exclusive switching, I'm sure there is someone out there using both the dogstatsd and the librato client …

glightfoot:
Made the change - is this the behavior you were expecting?

```
[tag1=bar,tag2=bazfoo.test:100|c
 &event.CounterEvent{CMetricName:"[tag1=bar,tag2=bazfoo.test", CValue:100, CLabels:map[string]string{}}

tag1=bar,tag2=baz]foo.test:100|c
&event.CounterEvent{CMetricName:"tag1=bar,tag2=baz]foo.test", CValue:100, CLabels:map[string]string{}}

foo.tag1=bar,tag2=baz]test:100|c
&event.CounterEvent{CMetricName:"foo.tag1=bar,tag2=baz]test", CValue:100, CLabels:map[string]string{}}
```

matthiasr:
This is great, thank you!

PR #317, opened 2020-06-19
Changelog and documentation update for #314

matthiasr:
Update the README to reflect #314. Add an entry to the changelog
with extended compatibility notes.

@glightfoot did I misrepresent anything?


glightfoot:
This is great! Thanks for the write up. The only thing I can think of is the change to statsd metric labels from "timing" to "observer" might be worth calling out in the changelog too.

PR #329, opened 2020-08-10
add lifecycle api

### LINKED ISSUE #327: Support reloading configuration from POST
To make running statsd-exporter simpler in k8s, it would be nice if we could reload the mappings via http POST, like prometheus does. I propose adding a web-admin api with the same flags and path as prometheus to support this

glightfoot:
Fixes #327. Adds a lifecycle API with the same semantics as Prometheus. Moves the http serving to use a mux, and also moves starting the HTTP server to after the mapper is initialized.

@matthiasr

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ Also, tags without values (`#some_tag`) are not supported and will be ignored.
 NOTE: Version 0.7.0 switched to the [kingpin](https://github.com/alecthomas/kingpin) flags library. With this change, flag behaviour is POSIX-ish:
 
 * long flags start with two dashes (`--version`)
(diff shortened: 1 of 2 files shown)


jimdaga:
Sorry to follow up on a closed PR, but is an empty reply expected? It results in non-zero exit code when using curl.
```
$ curl -X POST  http://localhost:9102/-/quit
curl: (52) Empty reply from server

$ echo $?
52
```

I'd be happy to open an Issue if that's preferred.

matthiasr:
Yes, please do!

WojciechKuk:
is it restarting counters/gauges from the scratch, or only updates mappings?

Issue #330, opened 2020-08-25
Registry as an interface?


bakins:
I want to embed statsd_exporter in another project rather than copy-paste large portions of it.  I do not want to use the default prometheus registery.  Thoughts on the Registry type used by Exporter being an interface?  It'd have a more functions than I normally like to have for an interface, but not too terrible.   The current registry implementation would be the only supported/in-tree implementation.

I can do a PR if you think the idea is acceptable.

matthiasr:
Yes please, this is perfectly in line with my thoughts on splitting the exporter up into reusable packages and re-using them in other places. The next use case I have in mind is the Graphite exporter.

PR #331, opened 2020-08-26
Add option to allow using another Registerer

bakins:
Possible different approach to my need in https://github.com/prometheus/statsd_exporter/issues/330

This solves my immediate issue of not using default Registerer.  This could be done in addition to making Registry an interface.

Thoughts?


matthiasr (commented):
This approach looks good to me. Could you add documentation comments to the NewRegistry and WithRegisterer functions?

matthiasr:
Once this is in, should I make a release? Still unsure about go modules …

bakins:
Closed in favor of #343

PR #333, opened 2020-09-04
Use prometheus.Gauge interface instead of prometheus.Counter

shmsr:
Well, it doesn't change any behaviour because `prometheus.Gauge` implements `prometheus.Counter` and the conversion works also fine.

But it'd better that we fix this by using `prometheus.Gauge` for `gauge`.


matthiasr (approved):
Good catch!

PR #334, opened 2020-09-06
Error strings shouldn't be capitalized 

shmsr:
The error strings shouldn't be capitalized as it's considered a good practice. Because if look up the function call stack, then the function using that error message won't like if the string is capitalized.

Also, now it matches the format as the other error strings (i.e., not capitalized) present in the codebase.


matthiasr:
Thank you!

PR #335, opened 2020-09-06
Avoid double initialization of variables

shmsr:
This PR fixes the case when the metric is not present in `rm` (registered metric) then `LastRegisteredAt` and `TTL` are initialized twice. Firstly, `LastRegisteredAt` is initialized with the `time.Time`'s zero value and then with `clock.Now()`. Plus, `TTL` is initialized with the same value twice.


matthiasr:
This is indeed cleaner, thank you!

shmsr:
Thanks, @matthiasr!

Issue #338, opened 2020-09-10
Add isAlive endpoint for k8s deployments


yaron-idan:
I'm deploying the exporter in our k8s cluster using this [chart](https://github.com/niclic/helm-charts/tree/master/prometheus-statsd-exporter), which probes the `/metrics` to make sure the exporter is alive.

I would rather use a dedicated `/isAlive` endpoint for this matter, any objections for adding one?
If this seems like a good idea, I might be able to create a small PR to add that.

matthiasr:
Please use the same endpoints as the [Prometheus management API](https://prometheus.io/docs/prometheus/latest/management_api/) for consistency.

matthiasr:
Fixed in #339

PR #339, opened 2020-09-12
add healthcheck endpoint

yaron-idan:
Closing issue #338

### DIFF
--- a/main.go
+++ b/main.go
@@ -15,6 +15,7 @@ package main
 
 import (
 	"bufio"
+	"fmt"
 	"net"
 	"net/http"
 	_ "net/http/pprof"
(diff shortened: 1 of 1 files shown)


yaron-idan:
Hey @matthiasr, can I please get this reviewed?

matthiasr:
Sorry about that, I was on vacation for a while and then the first week of work was wild.

For these simple health checks, I don't think we need a flag to turn them off. They don't expose anything, and who doesn't need them is free to ignore them. On the other hand, for consistency with prometheus/prometheus, would it make sense to also have a "ready" endpoint? When would that become active?

yaron-idan:
Hey @matthiasr , thanks for your comments!
I see your point about the flag necessity, I've pushed a commit removing it.
Regarding the "ready" endpoint - contrary to prometheus, which requires it's TSDB backend to become ready before it can successfully serve queries and scrape targets, this exporter can receive metrics the moment it's webserver is able to listen for requests. That's why I thought it might be redundant to add it as well. I've added this endpoint as well, but for now it fully replicates the healthy endpoint (only returning a different text).
If you think there are further checks that should be done before the service is ready let me know and I'll try to dig into that.

matthiasr (approved):
Looks good, having them do the same thing is a good start. If we find situations where they should differ, we can add more checks. Can you add the DCO sign-off please?

PR #341, opened 2020-09-28
Minor documentation improvements

dsabsay:
@matthiasr The README for this exporter is so well-written and informative that I was compelled to improve a couple _minor_ flaws.

* Fix CLI help for `--statsd.event-flush-interval`. I made an educated guess here based on a cursory reading of the code. If someone could double-check that I interpreted it correctly, that'd probably be wise.
* Move verbiage regarding default histogram bucket values to proper section.


matthiasr:
Thank you!

PR #347, opened 2020-11-20
Pass around custom registry for registering exporter metrics

rfratto:
Importers of pkg/exporter may not want for series to be imported into the default registry. This PR makes it possible to provide a custom registry for metrics registration.

Signed-off-by: Robert Fratto <robertfratto@gmail.com>


matthiasr (approved):
Ah, the joys of dependency injection 😂  looks good, and a good idea to allow this, thank you!

PR #352, opened 2020-12-10
Use a Parser interface to allow alternate implementations

bakins:
Allows other implementations of the parser for projects that reuse much of statsd_exporter. I'm using a custom format, but https://github.com/prometheus/graphite_exporter could possible use this as well.


matthiasr:
Looks great, thank you!

PR #354, opened 2020-12-18
Reorder and rework glob vs. regex documentation

### LINKED ISSUE #349: glob takes precedence over regex, bug or design?
I'm sending `job.fetch_duration_seconds:1.000000|ms` to a `statsd_exporter` running with below's mapping config:
```
mappings:
- match: ([^.]+)\.([^.]+)
  match_type: regex
  name: $2
  observer_type: histogram
  labels:
    job: $1
- match: "*.*"
  name: $2
  labels:
    job: $1
```
The metric line is caught by the second (`glob`) rule, not the first one (`regex`), and the `summary` is created (not the `histogram` - as I would expect).

Is it a bug or design that `glob` rules always takes precedence over `regex` ones?

matthiasr:
Note that regular expression matches are only evaluated after glob
matches. Add headings and introductory sentences to each glob type.

Remove the technical reasoning for choosing glob vs. regex; instead
explain the performance implications and gotchas of each type in turn.

Closes #349.

@faja does this clarify things for you?

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,15 +240,21 @@ mappings:
 ```
 
 Glob matching offers the best performance for common mappings.
-There are however pathological cases like the following matches:
 
-    a.*.*.*.*
-    a.b.*.*.*
-    a.b.c.*.*
-    a.b.c.d.*
+#### Ordering glob rules
 
-Optimize these mappings by reversing the order, or by disabling mapping ordering.
+List more specific matches before wildcards, from left to right:
+
+    a.b.c
+    a.b.*
+    a.*.d
+    a.*.*
+
+This avoids unexpected shadowing of later rules, and performance impact from backtracking.
+
+Alternatively, you can disable mapping ordering altogether.
 With unordered mapping, at each hierarchy level the most specific match wins.
+This has the same effect as using the recommended ordering.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -234,6 +239,26 @@ mappings:
...
+Glob matching offers the best performance for common mappings.
+There are however pathological cases like the following matches:
+
+    a.*.*.*.*
+    a.b.*.*.*
+    a.b.c.*.*
+    a.b.c.d.*
+
+Optimize these mappings by reversing the order, or by disabling mapping ordering.
+With unordered mapping, at each hierarchy level the most specific match wins.
```
SuperQ, line 251: This is confusing. Is the example order optimal or is the reverse of the example optimal?

fffonion: It'll be optimal either if we turn off ordering (set `glob_disable_ordering` to true, which in case user doesn't care about order at all), or keep `glob_disable_ordering` being false but reorder the rules in reverse order.

matthiasr: thanks, I changed it to make a positive recommendation instead of a negative example. This all makes me think we should maybe take away some options and simplify things …

PR #357, opened 2021-01-20
Refactor startup order to not grab ports when checkConfig is true

alexanderritola:
This moves the logic pertaining to setting up listeners after the exporter is created and the `checkConfig` (#263) early exit is evaluated.

In a CI setting, where the goal is to validate the config, setting up listeners seems unnecessary and presents opportunities for unrelated build failures.

I tested a number of startup configurations and I don't believe this breaks anything.

@matthiasr

### DIFF
--- a/main.go
+++ b/main.go
@@ -295,19 +295,47 @@ func main() {
 
 	cacheOption := mapper.WithCacheType(*cacheType)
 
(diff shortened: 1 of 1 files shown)


matthiasr (approved):
Good call, thank you very much!

PR #359, opened 2021-01-26
Allow setting defaults for histogram and summary options

### LINKED ISSUE #336: Missing support for default summary and histogram options
I'm able to configure default values for quantiles in the configuration file under `defaults`. However, `max_summary_age` and `summary_age_buckets` seem to be only defined within `mappings` for a specific metric. Is there a way to define these 2 parameters globally, for all metrics?

glightfoot:
This allows setting histogram_options and summary_options in the defaults section. I added some tests and refactored the mapper tests to use named subtests, but I still want to test it more thoroughly before I mark this as ready.

Feedback appreciated.

Fixes #336 @matthiasr

### DIFF
--- a/pkg/mapper/mapper.go
+++ b/pkg/mapper/mapper.go
@@ -22,8 +22,9 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/log"
-	"github.com/prometheus/statsd_exporter/pkg/mapper/fsm"
 	yaml "gopkg.in/yaml.v2"
+
+	"github.com/prometheus/statsd_exporter/pkg/mapper/fsm"
 )
 
 var (
@@ -100,7 +101,7 @@ func (m *MetricMapper) GetDefaultHistogramOptions() HistogramOptions {
 		r.Buckets = prometheus.DefBuckets
 	} else {
 		r = m.Defaults.HistogramOptions.Clone()
-		if m.Defaults.HistogramOptions != nil && len(m.Defaults.HistogramOptions.Buckets) == 0 {
+		if len(m.Defaults.HistogramOptions.Buckets) == 0 {
 			r.Buckets = prometheus.DefBuckets
 		}
 	}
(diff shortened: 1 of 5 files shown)


pkg/mapper/mapper_test.go
```diff
@@ -29,18 +29,23 @@ type mappings []struct {
...
 }
 
 func TestMetricMapperYAML(t *testing.T) {
 	scenarios := []struct {
+		testName  string
 		config    string
 		configBad bool
 		mappings  mappings
 	}{
 		// Empty config.
```
matthiasr, line 42: I think with the `testName` property we can remove the (now duplicate) comments, but it's also fine if we leave them in.

matthiasr (commented):
I like it, thank you for tackling it! Can you also update the [corresponding section in the README](https://github.com/prometheus/statsd_exporter#global-defaults)?

pkg/mapper/mapper_test.go
glightfoot: Done.

pkg/mapper/mapper.go
```diff
@@ -77,12 +100,16 @@ func (m *MetricMapper) InitFromYAMLString(fileContents string, cacheSize int, op
 		return err
 	}
 
-	if n.Defaults.Buckets == nil || len(n.Defaults.Buckets) == 0 {
-		n.Defaults.Buckets = prometheus.DefBuckets
+	if n.Defaults.HistogramOptions == nil {
```
glightfoot, line 103: I thought about moving this into the mapper_defaults file, as setting the defaults behavior seems like it could go there (if someone uses the defaults without this, it would be weird). Open to suggestions

glightfoot:
This ended up being a bigger rabbit hole than I thought initially...

pkg/mapper/mapper.go
```diff
@@ -70,19 +71,37 @@ var defaultQuantiles = []metricObjective{
 	{Quantile: 0.99, Error: 0.001},
 }
 
+// GetDefaultHistogramOptions returns the default HistogramOptions with default buckets if none are present
```
glightfoot, line 74: Do these functions need to make a copy of the default structs and return that since the registry uses these functions to get the options and then overwrites them?

glightfoot:
closing in favor of #361

PR #360, opened 2021-01-26
Don't return empty responses to lifecycle api requests

### LINKED ISSUE #358: Empty response from the /-/quit endpoint
**Describe the bug**
When using the new `/-/quit` endpoint to stop a sidecar container running **statsd_exporter** we use curl. 
However in the current implementation we get an empty response, resulting in a non-zero exit code (code: 52). 

**To Reproduce**
```
./statsd_exporter --web.enable-lifecycle
```
```
$ curl -X POST  http://localhost:9102/-/quit
curl: (52) Empty reply from server

$ echo $?
52
```

**Expected behavior**
A healthy response that results in a zero exit code/http 200.

**Your environment**
* Tested using `v0.19.0` 

**Additional context**
This is where the code was introduced: https://github.com/prometheus/statsd_exporter/pull/329 

I'm new to golang, so my attempts at a fix haven't worked; but I think something along the lines of `signal.Notify(signals, os.Interrupt, syscall.SIGTERM)` will be better than `os.Exit(0)`.

glightfoot:
Don't return empty responses to the lifecycle api and use the existing signals chan to quit instead of os.exit

Fixes #358 @matthiasr 

```
$ curl -X POST  http://localhost:9102/-/quit 
Requesting termination... Goodbye!%
$ echo $? 
0
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -477,9 +477,14 @@ func main() {
 			</body>
 			</html>`))
 	})
+
+	signals := make(chan os.Signal, 1)
+	signal.Notify(signals, os.Interrupt, syscall.SIGTERM)
+
 	if *enableLifecycle {
 		mux.HandleFunc("/-/reload", func(w http.ResponseWriter, r *http.Request) {
 			if r.Method == http.MethodPut || r.Method == http.MethodPost {
+				fmt.Fprintf(w, "Requesting reload")
 				if *mappingConfig == "" {
 					level.Warn(logger).Log("msg", "Received lifecycle api reload but no mapping config to reload")
 					return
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -490,8 +495,9 @@ func main() {
 		})
 		mux.HandleFunc("/-/quit", func(w http.ResponseWriter, r *http.Request) {
 			if r.Method == http.MethodPut || r.Method == http.MethodPost {
+				fmt.Fprintf(w, "Requesting termination... Goodbye!")
 				level.Info(logger).Log("msg", "Received lifecycle api quit, exiting")
-				os.Exit(0)
+				signals <- os.Kill
```
matthiasr, line 500: I'm curious how this works, since we do not `.Notify` on this signal in particular (IIUC it's not normally one that can be caught)? To me it would be clearer to send `os.Interrupt` here, would that have unintended side effects?

glightfoot: You are correct in that we only notify that channel on certain signals, but in this case it doesn't matter what signal we send to the channel, as anything sent to that channel will cause it to quit. I can add a quit channel if that would be a little clearer

glightfoot:
@matthiasr I modified it to use a quit channel with a select to be a little more clear.

main.go
matthiasr: ah, that makes sense, thank you!

matthiasr:
Awesome, thank you for fixing this!

PR #361, opened 2021-01-28
Allow setting defaults for histogram and summary options

### LINKED ISSUE #336: Missing support for default summary and histogram options

glightfoot:
The first PR got too complex as I found other little edge cases. This should be a simpler version.

This allows setting histogram_options and summary_options in the defaults section. I added some tests and refactored the mapper tests to use named subtests, but I still want to test it more thoroughly before I mark this as ready.

Feedback appreciated.

Fixes #336 @matthiasr

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -302,9 +302,9 @@ func (r *Registry) GetSummary(metricName string, labels prometheus.Labels, help 
 		}
 
 		summaryOptions := mapper.SummaryOptions{
-			MaxAge: r.Mapper.Defaults.SummaryOptions.MaxAge,
+			MaxAge:     r.Mapper.Defaults.SummaryOptions.MaxAge,
 			AgeBuckets: r.Mapper.Defaults.SummaryOptions.AgeBuckets,
-			BufCap: r.Mapper.Defaults.SummaryOptions.BufCap,
+			BufCap:     r.Mapper.Defaults.SummaryOptions.BufCap,
 		}
 
 		if mapping != nil && mapping.SummaryOptions != nil {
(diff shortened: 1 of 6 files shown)


pkg/registry/registry.go
```diff
@@ -295,14 +296,21 @@ func (r *Registry) GetSummary(metricName string, labels prometheus.Labels, help
...
 	if vh == nil {
 		metricsCount.WithLabelValues("summary").Inc()
-		quantiles := r.Mapper.Defaults.Quantiles
+		quantiles := r.Mapper.Defaults.SummaryOptions.Quantiles
 		if mapping != nil && mapping.SummaryOptions != nil && len(mapping.SummaryOptions.Quantiles) > 0 {
 			quantiles = mapping.SummaryOptions.Quantiles
 		}
-		summaryOptions := mapper.SummaryOptions{}
+
+		summaryOptions := mapper.SummaryOptions{
```
glightfoot, line 304: I'm not sure if it's strictly necessary to uses the defaults here, but this feels more correct since some people might use the registry without the mapper (I do this in a few projects). Users of the mapper or registry package will need to update their usage

matthiasr (approved):
This looks good! Let me know when you think it's ready to merge.

glightfoot:
I still haven't had time to actually integration test this, but I think it works as intended. I'll mark it ready for review in case someone else wants to test it, as the next few weeks are going to be busy for me.

matthiasr:
I gave it a quick run but I think the tests are pretty comprehensive for this 😄  thank you for implementing it!

PR #363, opened 2021-02-07
Better mapper cache interface

### LINKED ISSUE #295: Factor out mapper caches into their own package
This is a follow-up to #281. Part of #234.

The mapper package has taken on a lot of variation and code branches, especially with the choice to use no/LRU/RR caching.

"Cache the result of a mapping" is a separate thing from "map this to that". However, they both have the same inputs (event with statsd metric types + dot-notation metric + key/value tags) and outputs (Prometheus metrics type + underscore metric name + key/value labels). These may need to be formalized a bit but fundamentally we are already handling them.

Define an _interface_ that an initialized mapper should fulfill. Make the current mapper fulfill it. Write (based on the current caching code) several implementations that perform caching. They should be initialized with an "underlying" mapper (which could be the real mapper, a mock, or a completely different implementation).

Delete all caching related code from the mapper.

### LINKED ISSUE #362: License Issue for github.com/hashicorp/golang-lru
github.com/hashicorp/golang-lru has a Mozilla license, which is different from that of github.com/prometheus/statsd_exporter. 

The difference (and the mere presence of the former) makes it more difficult for open-source projects (e.g., https://fuchsia.dev) to vendor the latter.

License: MPL-2.0 license
More details at: https://raw.githubusercontent.com/spdx/license-list/master/MPL-2.0.txt

glightfoot:
Kind of fixes #295 and kind of fixes #362 by removing the lru code from the mapper package. 

First stab at moving this to a better interface. The noop cache is now default (was lru), and the rr and lru mapper caches have been moved to a different package. This makes it easier for users of the packages to implement their own caches as the interface is much simpler.

Instead of calling InitCache, there is a UseCache method on mapper to set a cache.

I still want to rework the metrics, but need to think about a better way to do so. I wanted to get some feedback before spending any more time on it.

@matthiasr 

Signed-off-by: glightfoot <glightfoot@rsglab.com>

### DIFF
--- a/pkg/mapper/mapper.go
+++ b/pkg/mapper/mapper.go
@@ -229,12 +229,10 @@ func (m *MetricMapper) InitFromYAMLString(fileContents string) error {
 	m.Defaults = n.Defaults
 	m.Mappings = n.Mappings
 
-	// If no cache has been configured, use a noop cache
-	if m.cache == nil {
-		m.cache = NewMetricMapperNoopCache()
-	}
 	// Reset the cache since this function can be used to reload config
-	m.cache.Reset()
+	if m.cache != nil {
+		m.cache.Reset()
+	}
 
 	if n.doFSM {
 		var mappings []string
@@ -266,6 +264,8 @@ func (m *MetricMapper) InitFromFile(fileName string) error {
 }
 
 func (m *MetricMapper) UseCache(cache MetricMapperCache) {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
 	m.cache = cache
 }
 
@@ -273,15 +273,13 @@ func (m *MetricMapper) GetMapping(statsdMetric string, statsdMetricType MetricTy
 	m.mutex.RLock()
 	defer m.mutex.RUnlock()
 
-	// default cache to noop cache if used from an uninitialized mapper
-	if m.cache == nil {
-		m.cache = NewMetricMapperNoopCache()
-	}
-
-	result, cached := m.cache.Get(formatKey(statsdMetric, statsdMetricType))
-	if cached {
-		r := result.(MetricMapperCacheResult)
-		return r.Mapping, r.Labels, r.Matched
+	// only use a cache if one is present
+	if m.cache != nil {
+		result, cached := m.cache.Get(formatKey(statsdMetric, statsdMetricType))
+		if cached {
+			r := result.(MetricMapperCacheResult)
+			return r.Mapping, r.Labels, r.Matched
+		}
 	}
 
 	// glob matching
@@ -303,13 +301,17 @@ func (m *MetricMapper) GetMapping(statsdMetric string, statsdMetricType MetricTy
 				Labels:  labels,
 			}
 			// add match to cache
-			m.cache.Add(formatKey(statsdMetric, statsdMetricType), r)
+			if m.cache != nil {
+				m.cache.Add(formatKey(statsdMetric, statsdMetricType), r)
+			}
 
 			return result, labels, true
 		} else if !m.doRegex {
 			// if there's no regex match type, return immediately
-			// Add miss cache
-			m.cache.Add(formatKey(statsdMetric, statsdMetricType), MetricMapperCacheResult{})
+			// Add miss to cache
+			if m.cache != nil {
+				m.cache.Add(formatKey(statsdMetric, statsdMetricType), MetricMapperCacheResult{})
+			}
 			return nil, nil, false
 		}
 	}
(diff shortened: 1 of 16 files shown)


glightfoot:
I just now realized the mapper cache issue was actually to wrap the mapper with a cache. I didn’t actually read that issue in detail, but I think having the mapper use a cache is important enough of a performance improvement to keep in there. Thoughts?

pkg/mapper/mapper.go
```diff
@@ -252,53 +256,34 @@ func (m *MetricMapper) InitFromYAMLString(fileContents string, cacheSize int, op
...
 
 func (m *MetricMapper) GetMapping(statsdMetric string, statsdMetricType MetricType) (*MetricMapping, prometheus.Labels, bool) {
 	m.mutex.RLock()
 	defer m.mutex.RUnlock()
-	result, cached := m.cache.Get(statsdMetric, statsdMetricType)
+
+	// default cache to noop cache if used from an uninitialized mapper
+	if m.cache == nil {
+		m.cache = NewMetricMapperNoopCache()
+	}
```
matthiasr, line 279: I think we need to take an exclusive lock here, not just an RLock?

matthiasr: I guess for performance _and_ correctness we need to check twice? (concurrency is not my strength)

```suggestion
	if m.cache == nil {
		m.mutex.Lock()
		if m.cache == nil {
			m.cache = NewMetricMapperNoopCache()
		}
		m.mutex.Unlock()
	}
```

pkg/mapper/mapper_benchmark_test.go
```diff
@@ -582,9 +585,18 @@ mappings:` + duplicateRules(100, ruleTemplateSingleMatchGlob)
...
 
 	for _, cacheType := range []string{"lru", "random"} {
+		mapper := MetricMapper{}
+		var cache MetricMapperCache
+		switch cacheType {
+		case "lru":
+			cache, _ = lru.NewMetricMapperLRUCache(mapper.Registerer, 1000)
+		case "random":
+			cache, _ = randomreplacement.NewMetricMapperRRCache(mapper.Registerer, 1000)
+		}
```
matthiasr, line 595: a bunch of tests are repeating this. would it be easier to have a helper that returns a (fresh) `map[string]MetricMapperCache` so we can `range cacheType, cache := cacheTypes() { … }`?

matthiasr: could we generalize things if that also includes the `noop` cache?

pkg/mapper/mapper_cache.go
```diff
@@ -56,155 +53,36 @@ func NewCacheMetrics(reg prometheus.Registerer) *CacheMetrics {
...
 	Labels  prometheus.Labels
 }
 
+// MetricMapperCache must be thread-safe and should be instrumented with CacheMetrics
 type MetricMapperCache interface {
-	Get(metricString string, metricType MetricType) (*MetricMapperCacheResult, bool)
-	AddMatch(metricString string, metricType MetricType, mapping *MetricMapping, labels prometheus.Labels)
-	AddMiss(metricString string, metricType MetricType)
+	// Get a cached result
+	Get(metricKey string) (interface{}, bool)
```
matthiasr, line 65: What is the benefit of this compared to keeping the metricString and metricType separate?

matthiasr (commented):
Aesthetically I would find the wrapping approach more pleasing, but this is definitely an improvement, and I am happy with it! I left a few comments but nothing major stands out.

pkg/mapper/mapper.go
glightfoot: good catch! yeah, we will need to check twice

pkg/mapper/mapper_cache.go
glightfoot: In my mind, it made more sense to build the key from the two parts and pass it into the cache already formatted. This way it makes the cache interface more generic (key-value, vs key1 key2 - value), and also keeps people from having to duplicate the key-building code in each cache package.

pkg/mapper/mapper_benchmark_test.go
glightfoot: yeah, I'll add a helper function for this.

pkg/mapper/mapper.go
glightfoot: It might make more sense to remove the noop cache entirely, and just hide the part that checks the cache for a result behind the `if m.cache == nil`

glightfoot:
@matthiasr I removed the noop cache entirely and hid the cache calls behind `if m.cache != nil` checks, which seems simpler. Also added a helper function for the tests

pkg/mapper/mapper.go
matthiasr: aaawwwe I liked it (again, aesthetically) but I don't feel strongly about this

matthiasr (approved):
LGTM! What is missing to take it out of Draft?

glightfoot:
Ahh sorry about that! Got busy with a job change. Should be ready for review

matthiasr:
Thanks a lot!

PR #365, opened 2021-02-07
Support metric name segments starting with numbers

### LINKED ISSUE #328: Support metric name segments starting with numbers
Encountered this issue while using graphite_exporter to convert https://github.com/airbnb/kafka-statsd-metrics2 metrics to Prometheus format.

[Metric name RE](https://github.com/prometheus/statsd_exporter/blob/master/pkg/mapper/mapper.go#L30) doesn't support segments starting with numbers. That makes metric names like `kafka.server.FetcherStats.brokerHost.hostname.brokerPort.9092.clientId.ReplicaFetcherThread-0-1.BytesPerSec.1MinuteRate.gauge` become an invalid metric name.

glightfoot:
Fixes #328 

Rework metricLineRE to support matching segments that start with a number. I'm not entirely sure that this is the exact use case, or that the tests duplicate the issue that @rayjcwu was describing, but it seems to work. The full regex used for matching glob patterns is below, and I think valid, but please double check. 
```
^(\*|[a-zA-Z_](-?[a-zA-Z0-9_])*)(\.\*|\.[a-zA-Z0-9_](-?[a-zA-Z0-9_])*)*$
```

The only other difference I think this change makes is that now `*` is valid , whereas previously I think it would have failed - is that in spec?

@matthiasr

### DIFF
--- a/pkg/mapper/mapper.go
+++ b/pkg/mapper/mapper.go
@@ -28,10 +28,14 @@ import (
 )
 
 var (
-	statsdMetricRE    = `[a-zA-Z_](-?[a-zA-Z0-9_])*`
-	templateReplaceRE = `(\$\{?\d+\}?)`
+	// The first segment of a match cannot start with a number
+	statsdMetricRE = `[a-zA-Z_](-?[a-zA-Z0-9_])*`
+	// The subsequent segments of a match can start with a number
+	// See https://github.com/prometheus/statsd_exporter/issues/328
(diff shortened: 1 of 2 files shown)


matthiasr (commented):
Thank you! At the moment we are only testing metric segments that are all-digits, could you also add a test case for `test.99bottles.myapp.*` or so?

matthiasr:
I think just `*` ought to work, actually, maybe this is also worth adding a spec?

PR #375, opened 2021-04-28
Use groupcache for LRU cache

bakins:
Addresses https://github.com/prometheus/statsd_exporter/issues/362

Use a simple wrapper around groupcache LRU cache.


matthiasr:
This looks great! I wonder if there is a performance difference?

matthiasr:
Ah, I see, the hashicorp package does the same thing, so this should be as good or better.

PR #378, opened 2021-06-03
Update dependencies and drop large go-kit update

howardjohn:
Pulling in the same changes as done in
https://github.com/prometheus/common/issues/255

Before, we import `6,979,266` LOC. With this change, it drops to `1,543,794`


matthiasr:
Neat, thank you!

Issue #380, opened 2021-06-16
Cannot map StatsD metrics with consecutive dashes in name


evandam:
Trying to reference an incoming StatsD metric with two dashes `--` in the name results in an error starting `statsd-exporter`. Replacing the dashes with an underscore results in the metric not being mapped, though.

If I had to guess, https://github.com/prometheus/statsd_exporter/blob/master/pkg/mapper/mapper.go#L33-L37 should be tweaked to allow `*` dashes.

Example metric `default-ubuntu-1804_kitchen.fabio--requests.count:0|c`

```
level=debug ts=2021-06-16T23:54:45.193Z caller=listener.go:73 msg="Incoming line" proto=udp line=default-ubuntu-1804_kitchen.fabio--requests.count:0|c
```

Mapping with underscores:

```yaml
mappings:
- match: "*.fabio__requests.count"
  name: "fabio_requests_count"
  help: "Fabio request count"
  labels:
    hostname: "$1"
```

Results in autogenerated metric mapping, not what we specify.

```bash
$ curl -s localhost:9102/metrics | grep 'requests_count'
# HELP default_ubuntu_1804_kitchen_fabio___requests_count Metric autogenerated by statsd_exporter.
# TYPE default_ubuntu_1804_kitchen_fabio___requests_count counter
default_ubuntu_1804_kitchen_fabio___requests_count 0
```

Mapping with dashes as seen in incoming metric:

```yaml
mappings:
- match: "*.fabio--requests.count"
  name: "fabio_requests_count"
  help: "Fabio request count"
  labels:
    hostname: "$1"
```

`statsd-exporter` logs:

```
$ statsd_exporter --statsd.mapping-config=mapping.yml --log.level=debug
level=info ts=2021-06-16T23:57:30.868Z caller=main.go:321 msg="Starting StatsD -> Prometheus Exporter" version="(version=0.21.0, branch=HEAD, revision=ef6627b9f05350d54cd3bfea5afe36617d7eb5a4)"
level=info ts=2021-06-16T23:57:30.869Z caller=main.go:322 msg="Build context" context="(go=go1.16.5, user=root@8ace135a0329, date=20210610-07:24:32)"
level=error ts=2021-06-16T23:57:30.870Z caller=main.go:340 msg="error loading config" error="invalid match: *.fabio--requests.count"
```

SuperQ:
Yea, I'm not sure why the dash is outside the character class matching.

matthiasr:
[...]

PR #381, opened 2021-06-17
allow multiple dashes in StatsD metric names

evandam:
See https://github.com/prometheus/statsd_exporter/issues/380

@matthiasr I don't do any Go development so not sure if there's anything else to consider here 😅 

Thanks!

### DIFF
--- a/pkg/mapper/mapper.go
+++ b/pkg/mapper/mapper.go
@@ -30,10 +30,10 @@ import (
 
 var (
 	// The first segment of a match cannot start with a number
-	statsdMetricRE = `[a-zA-Z_](-?[a-zA-Z0-9_])*`
+	statsdMetricRE = `[a-zA-Z_]([a-zA-Z0-9_\-])*`
 	// The subsequent segments of a match can start with a number
 	// See https://github.com/prometheus/statsd_exporter/issues/328
-	statsdMetricSubsequentRE = `[a-zA-Z0-9_](-?[a-zA-Z0-9_])*`
+	statsdMetricSubsequentRE = `[a-zA-Z0-9_]([a-zA-Z0-9_\-])*`
 	templateReplaceRE        = `(\$\{?\d+\}?)`
 
 	metricLineRE = regexp.MustCompile(`^(\*|` + statsdMetricRE + `)(\.\*|\.` + statsdMetricSubsequentRE + `)*$`)
--- a/pkg/mapper/mapper_test.go
+++ b/pkg/mapper/mapper_test.go
@@ -68,7 +68,7 @@ func TestMetricMapperYAML(t *testing.T) {
 mappings:
 - match: test.dispatcher.*.*.*
   name: "dispatch_events"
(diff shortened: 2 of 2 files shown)


matthiasr:
There is a test that specifically said "double dashes are illegal" – it is intended to exercise _a_ case that doesn't match this regex. Could you change it to use some other illegal character?

matthiasr:
Also, please add the DCO sign-off, instructions are behind the Details link

SuperQ:
It might also be useful to add a positive test like this: (line 1467)

```go
    {
      testName: "Config with multipel dashes",
      config: `mappings:
- match: "*.fabio--requests.count"
  name: "fabio_requests_count"
  help: "Fabio request count"
  labels:
    hostname: "$1"`,
      mappings: mappings{
        {
          statsdMetric: "test.fabio--requests.count",
          name:         "fabio_requests_count",
          labels: map[string]string{
            "hostname": "test",
          },
        },
      },
    },              },
```

mapping.yml
```diff
@@ -0,0 +1,5 @@
+mappings:
```
matthiasr, line 1: did you commit this file intentionally?

matthiasr:
Something I am wondering: what happens if we accept a statsd metric with `--` but have no mapping for it? Will it be transformed to a valid Prometheus metric?

mapping.yml
evandam: @matthiasr you're too fast man! Adding a new test like @SuperQ recommended, committed by accident 😇

evandam:
In my super quick testing it seems like it gets dropped if there's no mapping:

```
level=debug ts=2021-06-17T16:51:00.278Z caller=line.go:214 msg="Bad line from StatsD" line=X
level=debug ts=2021-06-17T16:51:00.290Z caller=listener.go:73 msg="Incoming line" proto=udp line=--.count:0|c
level=debug ts=2021-06-17T16:51:00.290Z caller=listener.go:73 msg="Incoming line" proto=udp line=
```

A mapping for `match: "--"` is invalid.

```
level=error ts=2021-06-17T16:52:24.339Z caller=main.go:340 msg="error loading config" error="invalid match: --"
```

It seems reasonable enough to me but it's up to you folks!

matthiasr:
Hmm, but for metrics with a single dash there is an automatic mapping. It will be very confusing to users if _some_ metrics are mapped automatically but others are not. Do you feel up to extending the [metric name escaping](https://github.com/prometheus/statsd_exporter/blob/master/pkg/mapper/escape.go) so that it handles this input?

evandam:
Hey @matthiasr,it sounds like you folks know better than me to get this over the line. I don't write Go either so maybe I'm not the best to handle this if there's more to it than is already done 😅

matthiasr:
I created a new issue for handling the default case: #389.

matthiasr (approved):
Thank you, and sorry for holding it up for so long!

PR #383, opened 2021-06-25
filter empty metric name

royeo:
Receiving an empty metric name results in an error in pulling the metric, such as

```sh
An error has occurred while serving metrics:

"" is not a valid metric name
```

### DIFF
--- a/pkg/exporter/exporter.go
+++ b/pkg/exporter/exporter.go
@@ -76,6 +76,11 @@ func (b *Exporter) Listen(e <-chan event.Events) {
 
 // handleEvent processes a single Event according to the configured mapping.
 func (b *Exporter) handleEvent(thisEvent event.Event) {
+	if thisEvent.MetricName() == "" {
+		level.Error(b.Logger).Log("msg", "metric name cannot be empty")
+		return
+	}
+
 	mapping, labels, present := b.Mapper.GetMapping(thisEvent.MetricName(), thisEvent.MetricType())
 	if mapping == nil {
 		mapping = &mapper.MetricMapping{}


pkg/exporter/exporter.go
```diff
@@ -76,6 +76,11 @@ func (b *Exporter) Listen(e <-chan event.Events) {
 
 // handleEvent processes a single Event according to the configured mapping.
 func (b *Exporter) handleEvent(thisEvent event.Event) {
+	if thisEvent.MetricName() == "" {
+		level.Error(b.Logger).Log("msg", "metric name cannot be empty")
```
matthiasr, line 80: we cannot log at Error level for every event. In many applications you would produce thousands of lines per second, overwhelming the log system and drowning out everything else.

There is already [code that should handle this](https://github.com/prometheus/statsd_exporter/blob/3b062c01c8cde2f06b64968109a62e7f9608f39a/pkg/exporter/exporter.go#L106-L110), can you find out why it is not working in your case?

matthiasr: Oh, I think I see it, that is only reached when there is an explicit mapping and _that_ results in an empty metric name.

Could you please consolidate the empty-name handling to be _after_ the mapping? An explicit mapping may still recover from an incoming event without a name, by specifying one explicitly. As for the mapping-present-but-name-empty case, please log at debug level and increment an error counter.

Please add a test or two as well.

matthiasr (changes requested):

pkg/exporter/exporter.go
royeo: I did not configure Mapping，so `present` is  false

matthiasr: Yes – the exporter should check for the empty name whether `present` or not, _after_ mapping. Checking the incoming event is not right, because when a mapping is there it might "fix" this. I want to preserve this option, since it allows handling some legacy inputs that cannot be handled otherwise.

matthiasr:
I'll close this due to inactivity. Please let me know if you (or anyone else) wants to pick it up again!

PR #392, opened 2021-09-01
Update transition scenarios

matthiasr:
Recommend running the exporter as a sidecar, now that the relay feature
makes this easy.

Re-order the overview section to put the final state first, then explain
transition and try-out options.


SuperQ (approved):
LGTM

PR #409, opened 2021-11-16
Set numeric user to comply runAsNonRoot k8s policy

### LINKED ISSUE #406: Container has runAsNonRoot and image has non-numeric user (nobody), cannot verify user is non-root
When `runAsNonRoot` is enabled in k8s the non-numeric user is not acceptable.

aleksandr-vin:
When in k8s, container has `runAsNonRoot` policy and image has non-numeric user (nobody), then the deployment will fail as it cannot verify user is non-root.

Closes #406

Signed-off-by: Aleksandr Vinokurov <aleksandr.vin@gmail.com>


matthiasr:
Thank you!

PR #410, opened 2021-11-17
Use the instance registry instead of the prometheus.registry

mattdurham:
Allowing the usage of the instance for these options. For the Grafana Agent this is preventing certain types of stats from being pushed. @matthiasr


matthiasr:
Thank you! I can see that this is a bug but I am struggling to formulate a changelog entry. @mattdurham could you suggest one that explains to users what problem this release would fix for them?

matthiasr:
Ah, do I understand correctly that in the exporter itself this worked incidentally (because the Registerer we use is the global registry), but it affected library users?

PR #414, opened 2021-12-02
feat: support inject global labels

shuxnhs:


matthiasr:
Thank you for your PR! Unfortunately, as per the [exporter guidelines](https://prometheus.io/docs/instrumenting/writing_exporters/#target-labels-not-static-scraped-labels), this is not a feature that we will add to this exporter. As lined out in the guideline, use target relabeling instead.

shuxnhs:
i know, but if your have many exporters in different regions, you can‘t use target relabeling to distinguish by the label. this is why i add this feature.

matthiasr:
Distinguish regions by adding an external label to the regional Prometheus servers.

shuxnhs:
ok, thank you～

PR #441, opened 2022-06-17
Fixing broken benchmark for mapper

pedro-stanaka:
## Summary

During an investigation of resource usage in a branch of mine I noticed we
have broken benchmark for BenchmarkGlobNoOrderingWithBacktracking.
This is the only benchmark that needed the logger and it was breaking because
we were not passing a logger to the MetricMapper.


matthiasr:
Thank you!

Issue #438, opened 2022-06-10
[docs] Document using regex to mark the original metric name as label


pedro-stanaka:
### Problem and feature request

I am in need to build a custom UI that pairs the original statsd metric name which people are sending from applications with the names they see in Prometheus/Grafana. Since undoing the process of translating metric names is not-trivial (not to say almost impossible) due to how open metric naming is on statsd-sphere I would like to have a possibility to have metrics tagged with their original names before translation.

### Solution proposal

We could introduce a feature flag (or some cli argument) that enables such feature and in the `github.com/prometheus/statsd_exporter/pkg/mapper/MetricMapper` we could tag the final metric with a label `original_name`.

matthiasr:
Sorry for the long silence, I had to think about and test this out, and then forgot about it.

I found that this _is_ already possible, in a very general way using `$0` in the mapping config, but only if you are using the `regex` match type. It must have been possible with glob matching when that was translated to regular expressions, but in the reimplementation as a state machine, we lost this.

Rather than hard-coding a label, I would prefer if we fix the use of `$0` as a substitution:

```yaml
# tmp/mapping.yml
mappings:
- match: "glob.*"
  name: "test_globbed"
  labels:
    part: "$1"
    whole: "$0"
- match: "re\\.([^.]*)"
  name: "test_regex"
  match_type: regex
  labels:
    part: "$1"
    whole: "$0"
```

```console
$ ./statsd_exporter --log.level debug  --statsd.mapping-config tmp/mapping.yml &
level=info ts=2022-06-18T10:29:47.002Z caller=main.go:291 msg="Starting StatsD -> Prometheus Exporter" version="(version=, branch=, revision=)"
level=info ts=2022-06-18T10:29:47.002Z caller=main.go:292 msg="Build context" context="(go=go1.18.2, user=, date=)"
level=info ts=2022-06-18T10:29:47.003Z caller=main.go:341 msg="Accepting StatsD Traffic" udp=:9125 tcp=:9125 unixgram=
level=info ts=2022-06-18T10:29:47.003Z caller=main.go:342 msg="Accepting Prometheus Requests" addr=:9102
$ echo 'glob.one:1|c' | nc localhost 9125
level=debug ts=2022-06-18T10:29:56.193Z caller=listener.go:136 msg="Incoming line" proto=tcp line=glob.one:1|c
$ echo 're.one:1|c' | nc localhost 9125
[...]

PR #442, opened 2022-06-17
Allow templating `$0` with glob matching

pedro-stanaka:
## Summary

This is my implementation for #438. I went with a naive approach here, I am always adding a new label at `exporter.go` if the metric has not labelset we create a new one and put the new label there.


.gitignore
```diff
@@ -6,3 +6,4 @@ dependencies-stamp
 /.tarballs
 *~
 /vendor
+.idea
```
matthiasr, line 9: Please add this to your global .gitignore, not every project

matthiasr (changes requested):
I don't want to make this the default for everyone, and I am uneasy about even hard coding the label value.

As I commented in #438, I found that this used to be possible by mapping the `$0` value into a label (and it still is, if you use the `regex` type match), so I would rather fix that for glob mappings than introduce a new behavior.

pedro-stanaka:
Thanks for the feedback on the issue #438 @matthiasr. I will close that now and try to work with the regex solution you presented. 👍

matthiasr:
Oops got confused 😅

Issue #443, opened 2022-06-18
`pkg/relay TestRelay_RelayLine/multiple_lines` is flaky


matthiasr:
This happened [multiple](https://app.circleci.com/pipelines/github/prometheus/statsd_exporter/740/workflows/292e54ea-fff1-4b64-81d2-0ce7c266c0cf/jobs/2054) [times](https://app.circleci.com/pipelines/github/prometheus/statsd_exporter/742/workflows/fd54eb46-48a8-4f4c-8478-d2c58c63d2cc/jobs/2059) [recently](https://app.circleci.com/pipelines/github/prometheus/statsd_exporter/744/workflows/ad16e391-ed99-4e72-a0ff-bc24dded2720/jobs/2066):

```log
=== Failed
=== FAIL: pkg/relay TestRelay_RelayLine/multiple_lines (0.01s)
    udp.go:85: At: /home/circleci/project/pkg/relay/relay_test.go:67
    udp.go:115: Expected to find: "foo5:100|c|#tag1:bar,#tag2:baz\n"
    udp.go:116: But got: ""
    --- FAIL: TestRelay_RelayLine/multiple_lines (0.01s)

=== FAIL: pkg/relay TestRelay_RelayLine (0.01s)
```

When re-running, it usually works.

SandroJijavadze:
Wanted to fix this but couldn't figure out how.
Figured out when it happens though, and it happens in cases when channel select hits `case <-relayInterval.C: ` first instead of `case b := <-r.bufferChannel:`.
Although nothing is sent in relayInterval if buffer is empty, for some reason udp-testing package receives "".

https://github.com/prometheus/statsd_exporter/blob/848212e028a5f80c55f1db114dc0fa6ec33a3300/pkg/relay/relay.go#L101-L135

SandroJijavadze:
Related issue: https://github.com/golang/go/issues/8869

PR #445, opened 2022-06-22
fix flaky test gosched hack

SandroJijavadze:
Fixes flaky test #443  using `runtime.Gosched()` to allow other goroutine to receive from buffer.


matthiasr (approved):

matthiasr:
Thank you!

PR #447, opened 2022-06-29
Adding documentation about special regex match group

pedro-stanaka:
## Summary

This PR documents some doubts I had in the issue #438 and therefore closes it.
By using the approach I documented here, I was able to achieve the behavior
I described in #438.


matthiasr:
Thank you!

PR #461, opened 2022-08-26
Counter with histogram suffix will clash with histogram

pedro-stanaka:
## Summary

While running the exporter in production a colegue faced an error when a counter clashes with a previously collected histogram.
I managed to reproduce the bug with some tests cases and this PR fixes the problem.

## Bug description

The problem can be reproduced in the following manner.

1. A sample `foo:1|h` (a histogram - to be translated to a prometheus histogram) is sent to the exporter.
2. The exporter will register `foo_count`, `foo_sum` and `foo_bucket`.
3. Later a sample `foo_count:1|c` arrives at the exporter, the exporter will also register the counter.
4. The gatherer will detect that this counter is not part of the histogram, but a separate counter and will crash.

## Changes

- Fix the described bug, by checking for existing histograms before registering the counters.
- Improve isolation of test case `TestConflictingMetrics()` by using a new Prometheus registry for each test run.

### DIFF
--- a/pkg/exporter/exporter_test.go
+++ b/pkg/exporter/exporter_test.go
@@ -520,10 +520,11 @@ mappings:
 				events <- s.in
 				close(events)
 			}()
-			ex := NewExporter(prometheus.DefaultRegisterer, testMapper, log.NewNopLogger(), eventsActions, eventsUnmapped, errorEventStats, eventStats, conflictingEventStats, metricsCount)
+			registerer := prometheus.NewRegistry()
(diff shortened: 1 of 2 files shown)


matthiasr (commented):
This is great, thank you! Do we need the same check for gauges?

pedro-stanaka:
> This is great, thank you! Do we need the same check for gauges?

Good catch. Added a few more test cases and I am checking now for gauges as well.

pedro-stanaka:
From what I see, the rest should be covered.

pedro-stanaka:
Just checking, anything else needed here @matthiasr ?

matthiasr:
Sorry, my bad, I'm underwater with work. Will review as soon as possible.

matthiasr (approved):
Thank you!

PR #469, opened 2022-09-21
Print version, help etc to stdout

don-philipe:
The default behavior of kingpin lib is to print stuff to stderr. This is now changed by using the os.Stdout writer. Now information like version and help text will be printed to stdout.


matthiasr (approved):
Thank you!

PR #474, opened 2022-11-11
Update client_golang to enable native histograms

pedro-stanaka:
## Summary

As of client_golang 1.14+ we have support for sparse histograms. This will start to pave the way to allow the exporter not to need bucket configuration per metric anymore.
On top of that, the current behavior is that buckets are still used in the text exposition format, and only when protobuf is negotiated the new native histograms are used, so it should be safe to upgrade right now without changes for older version of prometheus.

### DIFF
--- a/README.md
+++ b/README.md
@@ -403,6 +403,8 @@ mappings:
   observer_type: histogram
   histogram_options:
     buckets: [ 0.01, 0.025, 0.05, 0.1 ]
+    native_histogram_bucket_factor: 1.1
+    native_histogram_max_buckets: 256
   name: "my_timer"
   labels:
     provider: "$2"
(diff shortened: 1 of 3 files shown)


matthiasr:
Could you please rebase? That should trigger a new CircleCI build, the previous failure was due to [an issue on their end](https://status.circleci.com/incidents/3d31fnpthj7y). You can drop the first commit, the library update came in with #476 already.

pedro-stanaka:
CircleCI still won't work 😦

matthiasr:
Strange. Can you try logging into CircleCI with GitHub? That ~should~ might fix its access to your fork

matthiasr:
I pushed your commits into a branch in the main repository, and that seems to have triggered something …

matthiasr:
Alright the tests that that triggered passed, that's good enough for me. Thank you for adding this!

Issue #481, opened 2023-01-11
Support prefixed glob matches


matschaffer-roblox:
I'm transition from https://docs.victoriametrics.com/vmagent.html to https://github.com/prometheus/graphite_exporter and ran into a case where I'm using prefixed globs to help distinguish from one of a few possible metric prefix cases.

Here's a simple example match:

```
- match: "prefixed_glob.prefix_*.*"
  name: prefixed_glob
  labels:
    prefix: "${1}"
    metric: "${2}"
```

This will fail with:

```
invalid match: prefixed_glob.prefix_*.*
```

I think I could maybe switch to regex matches, but I'll need 10 of them so I'm a little concerned with the possible overhead vs glob matching.

It'd be great if the statsd exporter mapper could handle prefixed globs, or if you suspect 10 regex matches is no big deal, let me know :)

matthiasr:
I think this would be a worthwhile feature, although I am not sure how to do it with the efficient state machine. If anyone wants to tackle this please do!

Meanwhile, unless you have very high throughput, I expect ~10 regex matches to work. Instead of deploying the exporter like an agent, consider deploying it as a sidecar. This makes it so that each instance of the expirter only needs to handle the metrics from one instance of the application. Additionally, you get per instance metrics "for free" – from the outside it stops mattering whether there are two processes communicating over some protocol, or one natively instrumented one.

matschaffer-roblox:
> have very high throughput

The data source is spark, so I suspect the total number of series is fairly high. Do you know offhand if graphite_exporter already has a metric I could track for match rate or processing time?

> Instead of deploying the exporter like an agent, consider deploying it as a sidecar

Good point. I'm doing that already which should limit the number of series depending on number of spark executors (which varies proportional to VM size).

PR #482, opened 2023-01-11
Prefixed glob matches

matschaffer-roblox:
This is just a setup for https://github.com/prometheus/statsd_exporter/issues/481 which allows prefixed globs to reach the state machine.

The hard work of actually supporting them still needs to get done before this can be proposed for merge.


matschaffer-roblox:
Closing this down since we're on a track toward doing jmx rather than graphite collection (for spark).

PR #484, opened 2023-01-27
Fix signalfx tag doc link

attenuation:
The origin link was 404, replace it with GitHub SignalFx-agent readme


matthiasr:
Thank you!

PR #491, opened 2023-03-13
Making mapperConfigDefaults public

gaantunes:
This is making the `mapperConfigDefaults` and  `metricObjective` structs inside `MetricMapper` object public, so that Grafana Agent or any other application embedding this exporter can instatiate `MetricMapper` without using the `InitFromYAMLString` directly.

This is really helpful when you are not defining your configuration using YAML standard.

### DIFF
--- a/pkg/mapper/mapper.go
+++ b/pkg/mapper/mapper.go
@@ -43,7 +43,7 @@ var (
 
 type MetricMapper struct {
 	Registerer prometheus.Registerer
-	Defaults   mapperConfigDefaults `yaml:"defaults"`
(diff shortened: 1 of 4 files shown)


matthiasr (approved):
Makes sense!

There seems to be a problem with CircleCI running the tests, is this because the `grafana` org denies it access?

gaantunes:
> Makes sense!
> 
> There seems to be a problem with CircleCI running the tests, is this because the `grafana` org denies it access?

I think it was just missing my account auth on CircleCI, I authenticated and it ran succesfully.

Issue #498, opened 2023-05-28
Feature Request: "scale" field in mapping config to convert units


ebracho:
For my use case it would be great if there was a `scale` field in the mapping config to allow converting metric values from non-base units (e.g. milliseconds, microseconds) to base units. [As per prometheus best practices](https://prometheus.io/docs/practices/naming/): 

> A metric name ...should use base units (e.g. seconds, bytes, meters - not milliseconds, megabytes, kilometers). 

Ex:

```yaml
mappings:
  - match: statsd.grpc_server.*.*.latency_us
    name: grpc_server_handling_seconds
    scale: 1e-6 # converts microseconds -> seconds
    labels:
      grpc_service: "${1}"
      grpc_method: "${2}"
```

PR #499, opened 2023-05-28
Add `scale` field to mapping config

ebracho:
Fixes https://github.com/prometheus/statsd_exporter/issues/498

This field allows configuring unit conversions in mappings. This is important for following [prometheus best practices](https://prometheus.io/docs/practices/naming/), which states that metrics should use base units (e.g. seconds, not milliseconds). For example:

```yaml
mappings:
- match: foo.latency_ms
  name: foo_latency_seconds
  scale: 0.001
- match: bar.processed_kb
  name: bar_processed_bytes
  scale: 1024
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -577,6 +577,23 @@ metrics that do not expire.
  expire a metric only by changing the mapping configuration. At least one
  sample must be received for updated mappings to take effect.
 
+### Unit conversions
+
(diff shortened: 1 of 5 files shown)


SuperQ:
Nice.

This needs a DCO sign-off. You can use `git commit -s --amend` to add it.

matthiasr:
Very nice! Only one more thing I would ask: could you please add a section to the README about this so others can find this feature? Bonus points if you include an example 😄

matthiasr:
Wonderful, thank you!

ebracho:
Thanks for maintaining this project! It's been super helpful for us during a short-notice metrics pipeline migration.

PR #510, opened 2023-07-10
Convert line payloads to string for debug logging

### LINKED ISSUE #497: Inconsistent debug logging of line payloads from TCP listener
We log raw packets in different places. Sometimes, we pass it to the logger [as a string](https://github.com/prometheus/statsd_exporter/blob/c3752cf30f14d53838a3104c1fd1b20f3cc354eb/pkg/listener/listener.go#L75), sometimes [as a byte slice](https://github.com/prometheus/statsd_exporter/blob/c3752cf30f14d53838a3104c1fd1b20f3cc354eb/pkg/listener/listener.go#L136). Specifically, this affects the TCP ingestion path.

This leads to confusing log output, especially in the JSON format, because the "binary" data is Base64-encoded by the log library.

We should consistently convert the lines to strings for logging, since we know that it's mostly-ASCII data.

sumeshpremraj:
This avoids logging of bytes slices as Base64-encoded data, making it unreadable.

resolves #497


matthiasr (approved):
Nice, thank you!

PR #511, opened 2023-09-20
Performance improvement - process udp packets in separate goroutine in order to minimize packet drops

kullanici0606:
### Summary of this PR
In this PR I am introducing a queue for raw statsd udp payloads and a worker to process them in another goroutine

### Problem
As discussed in https://github.com/grafana/k6/issues/2044, when both number of packets arriving and size of the packets are large, statds_exporter starts to drop UDP packets because processing speed cannot catch the packet arrival speed

### Solution
In the past, I implemented similar UDP packet processing application in Java and best practice was there to read packet in one thread and process packets in another thread(s), therefore I introduced similar logic here: I added a queue / channel (chan []byte) so that the goroutine reading the packets from UDP socket can enqueue them in order to minimize time spent per packet and packets are processed in another goroutine which reads from the queue / channel

### Result
Packet drop rates decreased, most of the time no packet drops occur while in old version packets are always dropped for the case discussed in  https://github.com/grafana/k6/issues/2044

**Before:**
```
$ curl -s http://localhost:9102/metrics | grep udp
statsd_exporter_udp_packets_total 21270
```
`
packets in tcpdump (udp.dstport == 9125) 21477`

_207 packets are dropped_

**After:**
```
$ curl -s http://localhost:9102/metrics | grep udp
statsd_exporter_udp_packets_total 21384
```

`packets in tcpdump (udp.dstport == 9125) 21384`

_No packets are dropped_

**Future TODOs**
Memory benchmarks discussed in https://github.com/prometheus/statsd_exporter/pull/459

### DIFF
--- a/main.go
+++ b/main.go
@@ -71,6 +71,12 @@ var (
 			Help: "The total number of StatsD packets received over UDP.",
 		},
 	)
+	udpPacketDrops = promauto.NewCounter(
+		prometheus.CounterOpts{
+			Name: "statsd_exporter_udp_packet_drops_total",
+			Help: "The total number of dropped StatsD packets which received over UDP.",
+		},
+	)
 	tcpConnections = promauto.NewCounter(
 		prometheus.CounterOpts{
 			Name: "statsd_exporter_tcp_connections_total",
@@ -377,6 +383,7 @@ func main() {
 			Logger:          logger,
 			LineParser:      parser,
 			UDPPackets:      udpPackets,
+			UDPPacketDrops:  udpPacketDrops,
 			LinesReceived:   linesReceived,
 			EventsFlushed:   eventsFlushed,
 			Relay:           relayTarget,
--- a/main.go
+++ b/main.go
@@ -267,7 +267,7 @@ func main() {
 		signalFXTagsEnabled  = kingpin.Flag("statsd.parse-signalfx-tags", "Parse SignalFX style tags. Enabled by default.").Default("true").Bool()
 		relayAddr            = kingpin.Flag("statsd.relay.address", "The UDP relay target address (host:port)").String()
 		relayPacketLen       = kingpin.Flag("statsd.relay.packet-length", "Maximum relay output packet length to avoid fragmentation").Default("1400").Uint()
-		udpPacketQueueSize   = kingpin.Flag("statsd.udp-packet-queue-size", "Size of internal queue for processing udp packets.").Default("10000").Int()
+		udpPacketQueueSize   = kingpin.Flag("statsd.udp-packet-queue-size", "Size of internal queue for processing UDP packets.").Default("10000").Int()
 	)
 
 	promlogConfig := &promlog.Config{}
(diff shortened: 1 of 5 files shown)


pkg/listener/listener.go
```diff
@@ -64,12 +66,26 @@ func (l *StatsDUDPListener) Listen() {
 			level.Error(l.Logger).Log("error", err)
 			return
 		}
-		l.HandlePacket(buf[0:n])
+		// avoid making copies of slices since we need to minimize the time spent here in order not to drop packets
```
matthiasr, line 69: This comment is confusing to me – `EnqueueUdpPacket` immediately and synchronously goes on to make a copy of the packet?

pkg/listener/listener.go
```diff
@@ -64,12 +66,26 @@ func (l *StatsDUDPListener) Listen() {
...
+		l.EnqueueUdpPacket(buf, n)
 	}
 }
 
-func (l *StatsDUDPListener) HandlePacket(packet []byte) {
+func (l *StatsDUDPListener) EnqueueUdpPacket(packet []byte, n int) {
 	l.UDPPackets.Inc()
+	packetCopy := make([]byte, n)
+	copy(packetCopy, packet)
+	l.UdpPacketQueue <- packetCopy
```
matthiasr, line 78: What happens when the queue is full? Do we want to back pressure here, or would it be useful to drop the packet and increment a counter when this happens? For historical context, we [had this idea before](https://github.com/prometheus/statsd_exporter/pull/196#discussion_r274183349). The upside would be that it's easier to keep track of the dropped packets that way, instead of having to determine it from the kernel. At the time, we ended up not doing that in favor of other internal performance improvements.

Since it seems that you are still running into drops, I was wondering if this would be a good opportunity to revive that approach, especially since it _looks_ like in your single-processor approach it would be fairly easy to do this here?

pkg/listener/listener.go
matthiasr: If we really wanted to avoid allocations in the hot path, I guess we would need a rotating pool of buffers, but I'm not sure the memory overhead is worth it, so I would not go there right now. I think the best way here is to delete the confusing comment and accept that we _are_ making a copy.

main.go
```diff
@@ -261,6 +261,7 @@ func main() {
 		signalFXTagsEnabled  = kingpin.Flag("statsd.parse-signalfx-tags", "Parse SignalFX style tags. Enabled by default.").Default("true").Bool()
 		relayAddr            = kingpin.Flag("statsd.relay.address", "The UDP relay target address (host:port)").String()
 		relayPacketLen       = kingpin.Flag("statsd.relay.packet-length", "Maximum relay output packet length to avoid fragmentation").Default("1400").Uint()
+		udpPacketQueueSize   = kingpin.Flag("statsd.udp-packet-queue-size", "Size of internal queue for processing udp packets.").Default("10000").Int()
```
matthiasr, line 264:
```suggestion
		udpPacketQueueSize   = kingpin.Flag("statsd.udp-packet-queue-size", "Size of internal queue for processing UDP packets.").Default("10000").Int()
```

matthiasr (commented):
Thank you for this, and sorry for the delay, I was out for a few weeks and am still catching up.

Overall I really like the simplicity of this approach. I think it's ready to go with two minor edits. I am also wondering whether we should use this opportunity to handle the case where the queue is full explicitly, but I am happy to leave that to a future PR (or never, if it has undesirable side effects).

pkg/listener/listener.go
kullanici0606: Thank you for taking your time and reviewing my PR. You are right, comment is confusing, I removed the misleading comment. I also tried pooling but I think it made thinks worse since it needs some kind of syncronization.

pkg/listener/listener.go
kullanici0606: After your comment, I added a counter for UDP packet drops and incremented it when channel is full. Could you please review it?

PR #521, opened 2023-11-02
Add configuration for keeping original labels in mappings

rabenhorst:
When a label is assigned in mappings that already exists, it will be replaced. This PR introduces a `honor_labels` parameter. When this parameter is set to true, it will preserve the original value of the label.


matthiasr:
Awesome, thank you!

PR #541, opened 2024-02-28
Add global_labels option to defaults

erhudy:
global_labels provides a way to add labels that will be added to all published metrics from the exporter. If a label defined in global_labels is also defined in a mapping, it will be overridden.

This is something I set up for our internal use to help people who are using statsd bridge the gap to Prometheus metrics by being able to apply some static labels to all metrics being emitted from the exporter. It serves basically the same purpose as external_labels in the Prometheus server configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -471,8 +471,8 @@ mappings:
 
 ### Global defaults
 
-One may also set defaults for the observer type, histogram options, summary options, and match type.
-These will be used by all mappings that do not define them.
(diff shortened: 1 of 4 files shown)


matthiasr:
Thank you for your contribution! Unfortunately I have to say no to this – we have a [documented stance](https://prometheus.io/docs/instrumenting/writing_exporters/#target-labels-not-static-scraped-labels) that this should be done in scrape configuration, not by the exporter itself. I cannot undermine that in an official exporter 😅

Issue #557, opened 2024-05-22
[Feature Request] Support DogStatsD in Extended Aggregation mode


GrgDev:
### Problem

This tool has worked almost flawlessly for me. However, when using it to try to move one of my tenants from statsd to prometheus, we noticed that the timing metrics were all showing a flat line series when they shouldn't be. Long story short, we found out it was because they were using the extended aggregation config option to reduce network load. [See the docs on this mode here.](https://github.com/DataDog/datadog-go/blob/master/README.md#extended-aggregation)

In case that link gets broken later, here's the TL;DR description. Instead of sending:
```
my_distribution_metric:21|d|#all,my,tags
my_distribution_metric:43.2|d|#all,my,tags
my_distribution_metric:1657|d|#all,my,tags
```
it sends
```
my_distribution_metric:21:43.2:1657|d|#all,my,tags
```
This only gets applied to distributions, histograms, and timers. Counts and gauges still act as expected. It would be great if the exporter could handle this.

### Proposed Possible Solution

I'm a novice in Golang so take my suggestion with a grain of salt.

The `LineToEvents` function seems to be called on three different paths with all three being packet handling functions.
* [(l *StatsDUDPListener) HandlePacket(packet []byte)](https://github.com/prometheus/statsd_exporter/blob/master/pkg/listener/listener.go#L102)
* [(l *StatsDUnixgramListener) HandlePacket(packet []byte)](https://github.com/prometheus/statsd_exporter/blob/master/pkg/listener/listener.go#L217)
* [(l *StatsDTCPListener) HandleConn(c *net.TCPConn)](https://github.com/prometheus/statsd_exporter/blob/master/pkg/listener/listener.go#L168)
Since the extended aggregate metrics are multiple StatsD lines disguised as one line, I would assume they need to get expanded into multiple lines prior to entering that function. The two `HandlePacket` functions both have the same:
```
lines := strings.Split(string(packet), "\n")
```
at the start. We could replace that with a function that does the following:
* Split by newlines
* For each line, check if a line contains `|#`
  * If it does, check to see if there is more than one `:` to the left of the `|#`
    * If it does, copy the string as many times as there are `:` left of the `|#`
    * Replace the value section with only one value with each copied string getting one unique value from the list of values in the original line.
    * Replace the original line in the list of lines with the unpacked expanded list of lines without the extended aggregation.

[...]

PR #558, opened 2024-05-30
Add DogStatsD extended aggregation support

GrgDev:
This PR aims to add the ability for statsd exporter to receive [DogStatsD extended aggregation metrics](https://github.com/DataDog/datadog-go?tab=readme-ov-file#extended-aggregation). It is a feature I requested in Issue #557 and I wanted to try to implement it myself following the in-issue commented advice from @glightfoot. 

### Summary of Implementation

When DogStatsD has extended aggregation enabled, instead of sending:

```
my_distribution_metric:21|d|#all,my,tags
my_distribution_metric:43.2|d|#all,my,tags
my_distribution_metric:1657|d|#all,my,tags
```

it sends

```
my_distribution_metric:21:43.2:1657|d|#all,my,tags
```

This PR updates the `LineToEvents` function in the specific case of more than one `:` left of the initial `|`. If so, assume it is an extended aggregation line, convert it back into its multi-line non-aggregated form, then pass it onto the rest of the function as the samples. The logic enforces that this form of aggregation metrics must be only of the types of distribution, histogram, or timing.

### DIFF
--- a/pkg/line/line.go
+++ b/pkg/line/line.go
@@ -217,7 +217,7 @@ func (p *Parser) LineToEvents(line string, sampleErrors prometheus.CounterVec, s
 
 	labels := map[string]string{}
 	metric := p.parseNameAndTags(elements[0], labels, tagErrors, logger)
-	usingDogStatsDTags := strings.Contains(line, "|#")
+	usingDogStatsDTags := strings.Contains(elements[1], "|#")
 	if usingDogStatsDTags && len(labels) > 0 {
 		// using DogStatsD tags
 
--- a/pkg/line/line.go
+++ b/pkg/line/line.go
@@ -229,17 +229,34 @@ func (p *Parser) LineToEvents(line string, sampleErrors prometheus.CounterVec, s
 			return events
 		}
 
-		// handle DogStatsD extended aggregation
-		lineParts := strings.SplitN(elements[1], "|", 2)
+		lineParts := strings.SplitN(elements[1], "|", 3)
 		if strings.Contains(lineParts[0], ":") {
-			aggValues := strings.Split(lineParts[0], ":")
-			aggLines := make([]string, len(aggValues))
-			tags := lineParts[1]
-
-			for i, aggValue := range aggValues {
-				aggLines[i] = strings.Join([]string{aggValue, tags}, "|")
+			// handle DogStatsD extended aggregation
+			var isValidAggType bool
+			switch lineParts[1] {
+			case
+				"ms", // timer
+				"h",  // histogram
+				"d":  // distribution
+				isValidAggType = true
+			default:
+				isValidAggType = false
 			}
-			samples = aggLines
+
+			if isValidAggType {
+				aggValues := strings.Split(lineParts[0], ":")
+				aggLines := make([]string, len(aggValues))
+
+				for i, aggValue := range aggValues {
+					aggLines[i] = strings.Join([]string{aggValue, lineParts[1], lineParts[2]}, "|")
+				}
+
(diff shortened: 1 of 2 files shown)


pkg/line/line_test.go
```diff
@@ -403,6 +403,17 @@ func TestLineToEvents(t *testing.T) {
 		"datadog tag extension with both valid and invalid utf8 tag values": {
 			in: "foo:100|c|@0.1|#tag1:valid,tag2:\xc3\x28invalid",
 		},
+		"datadog timings with extended aggregation values": {
```
glightfoot, line 406: It might be good to add tests with metrics without tags as well, invalid tests where someone submits a line with a different tagging style (e.g. SignalFX) and the dogstatsd extended aggregation, and invalid tests for extended aggregation with counters and gauges

GrgDev: I believe I have now added these tests plus some extras as well as the additional logic required in the `line.go` code to pass them.

One more question. If parsing for DogStatsD tags is disabled, should a line with the extended aggregation format result in:
1. no tags but otherwise the same output as if DogStatsD parsing was enabled
2. no output as the line is considered a bad line
?

The other tests seemed to suggest that the main side effect of turning off parsing should be no tags, but since this is a DogStatsD specific feature, should this line be considered invalid altogether in such a scenario?

GrgDev:
@glightfoot Friendly nudge

pkg/line/line.go
```diff
@@ -217,18 +217,44 @@ func (p *Parser) LineToEvents(line string, sampleErrors prometheus.CounterVec, s
...
 
+		if isValidAggType {
+			aggValues := strings.Split(lineParts[0], ":")
+			aggLines := make([]string, len(aggValues))
+			_, aggLineSuffix, _ := strings.Cut(elements[1], "|")
+
+			for i, aggValue := range aggValues {
+				aggLines[i] = strings.Join([]string{aggValue, aggLineSuffix}, "|")
+			}
+			samples = aggLines
```
matthiasr, line 251: I'll accept this as is, but this amount of string manipulation is making me nervous – both for the potential to mess up a line, and because of the amount of intermediate strings we're leaving for the garbage collector to deal with.

I would _prefer_ if we could delay dealing with the multiple samples in a line until line 280, when we turn the value into a float. There, we could turn the single `value float64` into a `values []float64`, looping over the `:` separated values to parse them in turn. We should be able to do this without allocating any additional strings. Then, further down on line 331, where we're emitting `1/samplingFactor` events, we could also loop over that.

That being said, I've left this PR hanging for too long, so I'll take working over perfect. If anyone runs into actual issues from the current approach, or feels otherwise motivated, maybe this comment can help guide them 😃

matthiasr (approved):
Thank you for your contribution, and thank you @glightfoot for the initial review as well!

pkg/line/line.go
```diff
@@ -217,18 +217,44 @@ func (p *Parser) LineToEvents(line string, sampleErrors prometheus.CounterVec, s
...
+		level.Debug(logger).Log("msg", "Bad line (multiple tagging styles) from StatsD", "line", line)
+		return events
+	}
+
+	var samples []string
+	lineParts := strings.SplitN(elements[1], "|", 3)
+	if strings.Contains(lineParts[0], ":") {
+		// handle DogStatsD extended aggregation
+		isValidAggType := false
+		switch lineParts[1] {
```
m-barthelemy, line 235: After updating to 0.27.1, this line is triggering a `panic: runtime error: index out of range [1] with length 1' for us.

PR #579, opened 2024-09-15
Fix panic when parsing invalid lines

### LINKED ISSUE #572: panic in dogstatsd extended aggregation parsing
https://github.com/prometheus/statsd_exporter/pull/558#pullrequestreview-2259676259 – breaking this out into an issue by itself.

From @m-barthelemy:

> After updating to 0.27.1, [this line](https://github.com/prometheus/statsd_exporter/pull/558/files#diff-a9794bf96da36f26e5b9f52911da6346f9d53f8039ed1adf2e804b6782b04febR235) is triggering a panic: runtime error: index out of range [1] with length 1 for us.

Could you run the exporter with `--log.level=debug` and give us an example of a line that's triggering this? Ideally, add it as a test case and see if that also fails.

@GrgDev it seems that we need to check whether the splitting actually resulted in enough parts, do you have capacity to look into that?

matthiasr:
The line

```
|h|#consumer:Kafka::SharedConfigurationConsumer,topic:shared_configuration_update,partition:1,consumer_group:tc_rc_us
```

caused a panic because the line parsing _first_ splits by `:` and then failed to
find a `|` to split on.

Check that we get at least two "line parts" (i.e. splits around `|`) when we
expect them, and if not, gracefully reject the line instead of crashing.

Fixes #572.


matthiasr:
@GrgDev can you check my thinking here?

@m-barthelemy @twskipper if you have a chance, could you give this branch a spin to make sure it fixes the panic in your real world use cases, not just in unit tests?

pkg/line/line.go
```diff
@@ -229,6 +229,11 @@ func (p *Parser) LineToEvents(line string, sampleErrors prometheus.CounterVec, s
 
 	var samples []string
 	lineParts := strings.SplitN(elements[1], "|", 3)
+	if len(lineParts) < 2 {
+		sampleErrors.WithLabelValues("not_enough_parts_after_colon").Inc()
+		level.Debug(logger).Log("msg", "Bad line (not enough parts after first :) from StatsD", "line", line)
```
SuperQ, line 234:
```suggestion
		level.Debug(logger).Log("msg", "Bad line from StatsD, not enough parts after first ':'", "line", line)
```

matthiasr: No, at this point we have already split on `:` and _then_ we don't have enough `|` delimited parts left. I'll try to make the log message clearer.

matthiasr: The parsing code is a bit of a mess, partially because of the behavior it currently has where it needs to deal with all flavors at once (#580). It's hard to describe what's wrong with the line because we don't really know, we only know that we've chopped the line in half and now the second half doesn't look like we think it should.

SuperQ: Ahh, ok. Makes sense.

SuperQ (approved):

m-barthelemy:
> @GrgDev can you check my thinking here?
> 
> @m-barthelemy @twskipper if you have a chance, could you give this branch a spin to make sure it fixes the panic in your real world use cases, not just in unit tests?

I'd be happy to, but I don't know how to build an image from this project, had a quick look at the Dockerfile but it expects some already existing binary. Alternatively if there's an easy way for you to make a beta release image available on the Docker Hub, then it will be easy for me to test it on one of our non-production clusters.

matthiasr:
[Here you go](https://hub.docker.com/layers/prom/statsd-exporter/v0.27.2-pr579/images/sha256-0d0ccb2fb7b6f73b96368598e13b26c578c461dc2be26bcd5b822795d1c26f91?context=explore): `prom/statsd-exporter:v0.27.2-pr579`

m-barthelemy:
> [Here you go](https://hub.docker.com/layers/prom/statsd-exporter/v0.27.2-pr579/images/sha256-0d0ccb2fb7b6f73b96368598e13b26c578c461dc2be26bcd5b822795d1c26f91?context=explore): `prom/statsd-exporter:v0.27.2-pr579`

Thanks a ton!
Complete success so far:
- no more crashes
- can see `ts=2024-09-17T03:04:11.508Z caller=line.go:234 level=debug msg="bad line: not enough '|'-delimited parts after first ':'" ...` warnings being logged instead

m-barthelemy:
Thanks again for the fix! Will there be a new release and new Docker images soon?

matthiasr:
Yes! I was on vacation, sorry :)

matthiasr:
[v0.27.2](https://github.com/prometheus/statsd_exporter/releases/tag/v0.27.2) is out!

PR #587, opened 2024-10-21
Added 'collector.disable-go-runtime-metrics' flag

Amir-Hadadi:
This flag controls go runtime metrics collection by the exporter.
The default scenario is collecting the go runtime metrics using the default go Prometheus client, using this flag means we want to disable collection. It defaults to false.
Discussion and more info is [here](https://groups.google.com/g/prometheus-developers/c/NHs1Mg3oGXA).

### DIFF
--- a/main.go
+++ b/main.go
@@ -16,6 +16,7 @@ package main
 import (
 	"bufio"
 	"fmt"
+	"github.com/prometheus/client_golang/prometheus/collectors"
 	"net"
 	"net/http"
 	_ "net/http/pprof"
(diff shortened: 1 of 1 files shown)


Amir-Hadadi:
Subject about this is pending on Prometheus Developers

beorn7:
As discussed in the linked mail thread, I believe we should do this in a consistent way between exporters, ideally in https://github.com/prometheus/exporter-toolkit . (Note that this PR does something different to the existing flag in node_exporter, both in behavior and name of the flag.)

beorn7:
See also https://github.com/prometheus/exporter-toolkit/issues/119

matthiasr:
Thank you @beorn7 – I agree that this should be directly in exporter-toolkit, and consistent across exporters. I'll close this PR not as a "I don't want this" but as a "please do this in exporter-toolkit instead, and we can then pick it up here" 😄

PR #598, opened 2024-12-13
Update golang.org/x/crypto

cbartz:
because of CVE-2024-45337


SuperQ:
We use dependabot to automatically update these changes.

PR #605, opened 2024-12-23
Keep track of metric names on the metric statsd_exporter_events_conflict_total

pedro-stanaka:
## Summary

This is a modification we did to our internal version of the statsd-exporter and actually makes this metric actionable, so you can track down which metrics are problematic and fix them at the source (application code) or filter them out in the statsd exporter configuration.

If cardinality is a concern for other users, I can try enabling this behind a feature flag, and report a constant metric name if we dont want to track it.

### DIFF
--- a/main.go
+++ b/main.go
@@ -148,7 +148,7 @@ var (
 			Name: "statsd_exporter_events_conflict_total",
 			Help: "The total number of StatsD events with conflicting names.",
 		},
-		[]string{"type"},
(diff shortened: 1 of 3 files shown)


matthiasr (approved):
Hmm I _think_ cardinality here is bounded because if you had a cardinality explosion, they wouldn't be conflicting? So I'm good with this as-is

PR #611, opened 2025-01-29
Add basic TLS listener support

### LINKED ISSUE #413: Add TLS for the web UI + metrics
Use the [exporter toolkit](https://pkg.go.dev/github.com/prometheus/exporter-toolkit/web) to add support for TLS on the HTTP interface. See prometheus/graphite_exporter#175 for an example.

ag-TJNII:
This MR adds basic HTTPS listener support.  This was implemented using dedicated config arguments, same as listenAddress, and Go's `http.ListenAndServeTLS()`.  prometheus/exporter-toolkit was not used as the API of that project isn't clear from the documentation and per https://github.com/prometheus/exporter-toolkit/issues/197 the TLS support is slated for removal.

Default options are HTTP so this does not change existing behavior or require any config changes by users.

Resolves #413

### DIFF
--- a/main.go
+++ b/main.go
@@ -173,8 +173,15 @@ var (
 	)
 )
 
-func serveHTTP(mux http.Handler, listenAddress string, logger *slog.Logger) {
-	logger.Error(http.ListenAndServe(listenAddress, mux).Error())
+func serveHTTP(mux http.Handler, listenAddress string, listenTLSCert string, listenTLSKey string, logger *slog.Logger) {
+	if listenTLSCert == "" && listenTLSKey == "" {
+		logger.Info("Accepting Prometheus Requests over HTTP", "addr", listenAddress)
+		logger.Error(http.ListenAndServe(listenAddress, mux).Error())
+	} else {
+		logger.Info("Accepting Prometheus Requests over HTTPS", "addr", listenAddress)
+		logger.Error(http.ListenAndServeTLS(listenAddress, listenTLSCert, listenTLSKey, mux).Error())
+	}
+
 	os.Exit(1)
 }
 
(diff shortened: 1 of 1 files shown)


matthiasr (changes requested):
Thank you for kicking this off!

> per https://github.com/prometheus/exporter-toolkit/issues/197 the TLS support is slated for removal

No, that issue is not about removal of the feature, only refactoring how it is implemented.

> prometheus/exporter-toolkit was not used as the API of that project isn't clear from the documentation

This is true, but nonetheless we should not diverge from it in this project. It's not just about convenience of writing (and maintaining!) less code, it's also a common user interface across the Prometheus project, and it provides a common way to specify [many parameters](https://github.com/prometheus/exporter-toolkit/blob/master/docs/web-configuration.md) that do matter to many TLS users.

The [Graphite exporter `main()`](https://github.com/prometheus/graphite_exporter/blob/04e6da143c01cbaede1f005107235da44ca465d6/cmd/graphite_exporter/main.go#L76-L190) is a good starting point that illustrates how to set up the exporter-toolkit HTTP listener. The main steps are to [let it add the flags](https://github.com/prometheus/graphite_exporter/blob/04e6da143c01cbaede1f005107235da44ca465d6/cmd/graphite_exporter/main.go#L52), then [start the listener](https://github.com/prometheus/graphite_exporter/blob/04e6da143c01cbaede1f005107235da44ca465d6/cmd/graphite_exporter/main.go#L186). All the generic flags, including handling the details of which package implements the TLS configuration, are handled between these two.

ag-TJNII:
Ack, that's a reasonable rejection.  I'm going to close this and will reopen it later with requested changes.

ag-TJNII:
The level of refactor requested here is beyond what I can provide at this time.  The statsd_exporter uses a mux, whereas graphite does not, and statsd has several endpoints which graphite does not:  https://github.com/prometheus/statsd_exporter/blob/master/main.go#L513-L531  I will not be filing a followup PR at this time.

PR #642, opened 2025-07-17
Extract and modernize starter configuration

matthiasr:
Take the existing starter configuration that we had buried in the middle of the
README. Slightly modernize it: disable glob mapping ordering by default, and set
the default # of native histogram buckets to 160, matching [defaults in
OpenTelemetry and the recommendation in Mimir](https://grafana.com/docs/mimir/latest/configure/configure-native-histograms-ingestion/#configure-native-histograms-per-tenant).

Add an example of overriding the default native histogram bucket factor.

### DIFF
--- a/config.example.yml
+++ b/config.example.yml
@@ -1,6 +1,7 @@
 # This is a modern starter configuration for the statsd_exporter. It uses native
 # and classic histograms by default, as well as the fast glob matching without ordering.
 defaults:
+  # Convert timers into histograms, which can be aggregated across instances. See below for an example of using a summary with pre-computed quantiles instead.
   observer_type: histogram
   histogram_options:
     # Expose native histograms.
--- a/config.example.yml
+++ b/config.example.yml
@@ -1,0 +1,81 @@
+# This is a modern starter configuration for the statsd_exporter. It uses native
+# and classic histograms by default, as well as the fast glob matching without ordering.
+defaults:
+  observer_type: histogram
+  histogram_options:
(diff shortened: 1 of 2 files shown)


config.example.yml
```diff
@@ -0,0 +1,81 @@
...
+  histogram_options:
+    native_histogram_bucket_factor: 1.005
+    # For the sake of this example, set the buckets for a latency distribution
+    # tightly clustered around 15ms.
+    buckets: [0.01, 0.011, 0.012, 0.013, 0.014, 0.015, 0.016, 0.017, 0.018, 0.019, 0.02]
+  labels:
+    provider: "$2"
+    outcome: "$3"
+    job: "${1}_server"
+# Example 3: This will be a summary using the summary_options set in `defaults`
```
SuperQ, line 66:
```suggestion
# Example 3: This will be a summary using the summary_options set in `defaults`
#
# Note: Summary metrics are not recommended because they can not be aggregated over multiple instances.
```

config.example.yml
```diff
@@ -0,0 +1,83 @@
+# This is a modern starter configuration for the statsd_exporter. It uses native
+# and classic histograms by default, as well as the fast glob matching without ordering.
+defaults:
+  observer_type: histogram
```
matthiasr, line 4:
```suggestion
  # Convert timers into histograms, which can be aggregated across instances. See below for an example of using a summary with pre-computed quantiles instead.
  observer_type: histogram
```

PR #673, opened 2026-03-17
Drop deprecated fields + use `yaml.UnmarshalStrict()`

### LINKED ISSUE #546: Reject configuration with unknown fields

simonhammes:
This enables strict config parsing and removes all deprecated config fields (as suggested in https://github.com/prometheus/statsd_exporter/pull/608#discussion_r1916172092).

I also removed all test cases that turned out to be redundant.

Fixes #546


pedro-stanaka (approved):
please rebase against master

simonhammes:
> please rebase against master

Done, thanks for the review!

PR #684, opened 2026-04-15
pkg/line: support DogStatsD v1.3 metric timestamp field (|T<unix_ts>)

### LINKED ISSUE #680: DogStatsD v1.3 support an optional Unix timestamp field.
It appears that statsd_exporter may not support DogStatsD v1.3 https://docs.datadoghq.com/extend/dogstatsd/datagram_shell/?tab=metrics#dogstatsd-protocol-v13

docs:
The value is a Unix timestamp (UTC) and must be prefixed by T, for example:

<METRIC_NAME>:<VALUE>|<TYPE>|#<TAG_KEY_1>:<TAG_VALUE_1>,<TAG_2>|T<METRIC_TIMESTAMP>

[Example datagram](https://docs.datadoghq.com/extend/dogstatsd/datagram_shell/?tab=metrics#example-datagram)
page.views:15|c|#env:dev|T1656581400: A COUNT indicating that 15 page views happened on the 30th of June, 2022 at 9:30am UTC

Do you plan to add support for protocol 1.3 and possibly all others?

alliasgher:
## Summary

DogStatsD protocol v1.3 adds an optional timestamp suffix to the metric datagram:

```
<name>:<value>|<type>|#<tags>|T<unix_timestamp>
```

Previously any datagram containing this field was silently dropped because `T` was not a recognised component prefix (hit the `default` branch → `continue samples`) and the 5-component guard rejected fully-formed v1.3 datagrams.

## Changes

- **`pkg/line/line.go`**: add a `'T'` case in the component switch that validates the Unix timestamp integer and ignores it (statsd_exporter does not propagate source timestamps to Prometheus — same approach as the v1.2 `|c:<container_id>` field)
- Raise the max-components guard from `5` → `8` to accommodate all six possible v1.3 fields (`value`, `type`, `@sample`, `#tags`, `|c:`, `|T:`) without hitting a false `malformed_component` error
- **`pkg/line/line_test.go`**: five new test cases — counter, gauge, combined sampling+container+timestamp, and two malformed-timestamp cases

## Tests

```
go test ./pkg/line/...   # all pass
```

Fixes #680

### DIFF
--- a/pkg/line/line.go
+++ b/pkg/line/line.go
@@ -269,7 +269,7 @@ samples:
 	for _, sample := range samples {
 		samplesReceived.Inc()
 		components := strings.Split(sample, "|")
-		if len(components) < 2 || len(components) > 5 {
+		if len(components) < 2 || len(components) > 8 {
 			sampleErrors.WithLabelValues("malformed_component").Inc()
 			logger.Debug("bad component", "line", line)
 			continue
@@ -327,6 +327,20 @@ samples:
 					} else {
 						logger.Debug("Malformed container ID section", "component", component, "line", line)
 						sampleErrors.WithLabelValues("malformed_container_id").Inc()
+						continue samples
+					}
+				case 'T':
+					// Handle DogStatsD v1.3 metric timestamp (e.g., |T1656581400).
+					// The timestamp is a Unix UTC epoch; statsd_exporter does not forward
+					// source timestamps to Prometheus, so we validate the field and ignore it.
+					if len(component) < 2 {
+						logger.Debug("Malformed timestamp section", "component", component, "line", line)
+						sampleErrors.WithLabelValues("malformed_timestamp").Inc()
+						continue samples
+					}
+					if _, err := strconv.ParseInt(component[1:], 10, 64); err != nil {
+						logger.Debug("Invalid timestamp value", "component", component[1:], "line", line)
+						sampleErrors.WithLabelValues("malformed_timestamp").Inc()
 						continue samples
 					}
 				default:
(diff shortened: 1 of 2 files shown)


pedro-stanaka (commented):
Been busy times for me, finally had the chance to review this. Need some clarification and I think we can drop the expensive parse of the timestamp.

pkg/line/line.go
```diff
@@ -269,7 +269,7 @@ samples:
 	for _, sample := range samples {
 		samplesReceived.Inc()
 		components := strings.Split(sample, "|")
-		if len(components) < 2 || len(components) > 5 {
+		if len(components) < 2 || len(components) > 8 {
```
pedro-stanaka, line 272: We just added support for this new value, do we really need to bump to 8 here? What are the other things we are allowing passing through here?

pkg/line/line.go
```diff
@@ -329,6 +329,20 @@ samples:
...
+					if len(component) < 2 {
+						logger.Debug("Malformed timestamp section", "component", component, "line", line)
+						sampleErrors.WithLabelValues("malformed_timestamp").Inc()
+						continue samples
+					}
+					if _, err := strconv.ParseInt(component[1:], 10, 64); err != nil {
+						logger.Debug("Invalid timestamp value", "component", component[1:], "line", line)
+						sampleErrors.WithLabelValues("malformed_timestamp").Inc()
+						continue samples
+					}
```
pedro-stanaka, line 345: I dont think we need this, if we dont care about the data we shouldnt care if it is valid or not.

Issue #544, opened 2024-03-02
stackstorm mapping - match_metric_type not working right?


Beardedwan:
i'm trying to append metrics with their type because i'm trying to get my stackstorm metrics into prometheus using this exporter and the exposed metrics have both counters and timers with the same metric name (https://docs.stackstorm.com/reference/metrics.html#exposed-metrics).  So, i wrote a mapping file with the following: 

```
- match: "(.+)"
    match_type: regex
    match_metric_type: observer
    name: "${1}_seconds"

  - match: "(.+)"
    match_type: regex
    name: "${1}_counter"
    match_metric_type: counter

  - match: "(.+)"
    match_type: regex
    name: "${1}_gauge"
    match_metric_type: gauge
```

This works as expected, but ST2 exposes some more detail in a "fun" way of exposing detailed metrics... 
```
st2.action.\<action ref\>.executions | counter 
st2.action.\<action ref\>.executions | timer
```
where \<action ref\> contains (2) useful pieces of information i want to include as labels... so i tried 
```
  - match: "st2\\.action\\.(\\w+)\\.(\\w+)\\.executions(.*)"
    match_type: regex
    name: "st2_action_executions${3}"
    labels:
      st2_library: "${1}"
      st2_action: "${2}"`
```

This did exactly what i wanted but i lost the timer metrics entirely and started getting 
`"Failed to update metric" ....is already registered` issues.

SO, i thought, OK i'll break these up an leverage `match_metric_type` and specify the difference before the "catch alls" at the bottom of the mapper

``` 
- match: "st2\\.action\\.(\\w+)\\.(\\w+)\\.executions(.*)"
    match_type: regex
    metric_match_type: observer
    name: "st2_action_executions${3}_seconds"
    labels:
      st2_library: "${1}"
      st2_action: "${2}"

  - match: "st2\\.action\\.(\\w+)\\.(\\w+)\\.executions(.*)"
    match_type: regex
    metric_match_type: counter
    name: "st2_action_executions${3}_counter"
    labels:
      st2_library: "${1}"
      st2_action: "${2}"
```

but even with this... no luck... more "already registered" issues... so it seems like `match_metric_type` doesn't ALWAYS work right?

not quite the same as #355 because i can add the type to the metric name, but i can't seem to both add the type to the metric name AND create labels based on the metric name 

if i submit 
```
st2.action.fubar.pson_to_json.executions.scheduled.something:3.30400000|ms
[...]

PR #695, opened 2026-05-16
Reject unknown fields in mapping configuration

### LINKED ISSUE #546: Reject configuration with unknown fields

highlyavailable:
`InitFromYAMLString` used `yaml.Unmarshal`, which silently drops keys that don't match a struct field. A typo in a mapping or `defaults` key would be ignored, so the config quietly behaved differently than intended (see #544).

This switches to `yaml.UnmarshalStrict` so unknown keys are reported as an error. The deprecated `timer_type`, `buckets` and `quantiles` keys still work, since the alias structs the custom unmarshalers use already declare them. Added tests for an unknown key at the top level, in `defaults`, and in a mapping.

As noted in the issue this is technically a breaking change, in the sense that a config which silently did the wrong thing now fails loudly, so it probably wants a `[CHANGE]` changelog line at release.

Closes #546

### DIFF
--- a/pkg/mapper/mapper.go
+++ b/pkg/mapper/mapper.go
@@ -83,7 +83,7 @@ var defaultQuantiles = []MetricObjective{
 func (m *MetricMapper) InitFromYAMLString(fileContents string) error {
 	var n MetricMapper
 
-	if err := yaml.Unmarshal([]byte(fileContents), &n); err != nil {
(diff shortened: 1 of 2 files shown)


highlyavailable:
Closing this as a duplicate. I missed that #608, #673, and #687 already address this issue. Apologies for the extra noise.

PR #712, opened 2026-06-23
Reject configuration with unknown fields

### LINKED ISSUE #546: Reject configuration with unknown fields

Vedanshu7:
Fixes #546.

This switches `MetricMapper.InitFromYAMLString` from `yaml.Unmarshal` to `yaml.UnmarshalStrict`, so that mapping configs containing unknown or misspelled fields fail loudly instead of being silently ignored (the kind of problem seen in #544).

### Behavior change
This is potentially breaking: configurations that previously loaded while silently ignoring unrecognized fields will now return an error. Probably worth a `[CHANGE]` note in the changelog at release time.

### Tests
Added two cases to `TestMetricMapperYAML` covering an unknown top-level field and an unknown per-mapping field. Verified they fail without this change and pass with it; the full test suite passes.

cc @matthiasr @pedro-stanaka

### DIFF
--- a/pkg/mapper/mapper.go
+++ b/pkg/mapper/mapper.go
@@ -87,7 +87,7 @@ func (m *MetricMapper) InitFromYAMLString(fileContents string) error {
 		m.Logger = promslog.NewNopLogger()
 	}
 
-	if err := yaml.Unmarshal([]byte(fileContents), &n); err != nil {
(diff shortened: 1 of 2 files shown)


Vedanshu7:
Closing as a duplicate of #608, #673, and #687, which already address this. Apologies for the noise.

PR #714, opened 2026-06-25
Add lifecycle endpoint to clear metrics

### LINKED ISSUE #637: lifecycle web endpoint for clearing the registry
there are endpoints `/-/quit` and` /-/reload`
I report the need for the endpoint `/-/clear `(or `/-/reset`)

This command would work differently than reload. It would simply clear the registers (as if the TTL of all metrics dropped to zero).

I work with large "clusters" of statsd_exporter, where when the number of instances changes, or their order changes, the metrics distribution key changes (hashing on statsrelay).
In such a case I need a way for all statad_exporters to immediately "forget" the metrics they have accumulated so far, clear the registers and start collecting metrics "in a new distribution".
* /-/reload doesn't provide me with this - metrics don't get forgotten and new ones start to accumulate mixed with old ones
* /-/quit doesn't provide this because it's more disruptive. You can of course put a command in the systemd configuration to quickly start the daemon, but it can take a few seconds and it can cause prometheus to mark the backend as "down". I just want to clean the registry, which would be faster (and non-destructive for self-monitoring metrics i.e golang-related)

Other use cases that may appear (and user may need to quickly clean the registry):
- a wrongly formatted metric accidentally sent to statsd, which needs to be quickly cleaned up so that prometheus does not download it
- a gauge that is a memento of a non-existent instance of the monitored service - overstating/understating some KPI
- a mistake in metric routing (e.g. a mistake in redirecting traffic to the wrong backend on the loadbalancer, in a situation where all backends are downloaded equally by prometheus)

yuminn-k:
Fixes #637.

## Summary

This adds a `/-/clear` lifecycle endpoint, available only when `--web.enable-lifecycle` is enabled, alongside the existing `/-/reload` and `/-/quit` endpoints.

The endpoint clears dynamically registered StatsD time series without restarting the exporter. Internally, the clear request is handled by the exporter event loop, so it is serialized with normal event ingestion and stale metric cleanup. The registry clear path deletes registered label sets in the same way TTL expiration does, while keeping the existing vector/type metadata available for future samples.

cc @pedro-stanaka from the issue discussion.

## Verification

- `go test ./pkg/registry ./pkg/exporter -run 'TestClearMetrics|TestTtlExpiration|TestHashLabelNames' -count=1`
- `go test . -run TestClearMetricsHandler -count=1`
- `go test ./...`
- `go test -race ./pkg/exporter -run TestClearMetrics -count=1`
- `go test -race . -run TestClearMetricsHandler -count=1`
- `git diff --check`


pkg/exporter/exporter.go
```diff
@@ -73,6 +77,13 @@ func (b *Exporter) Listen(e <-chan event.Events) {
 	}
 }
 
+// ClearMetrics clears all dynamically registered StatsD time series.
+func (b *Exporter) ClearMetrics() int {
```
pedro-stanaka: This synchronously sends on an unbuffered channel and waits for a reply that is only processed inside `Listen()`. If called before `Listen()` starts, after it exits, or while it is stuck, callers can block indefinitely. This is especially risky now that `ClearMetrics` is public and callable by library users outside the binary's lifecycle flow.

**Suggestion:** Consider guarding send/receive with lifecycle state. A robust pattern is a `done` channel closed when `Listen` exits, then using `select` in `ClearMetrics()` for both send and receive paths (including `case <-done:`) so calls fail fast instead of hanging indefinitely.

pkg/exporter/exporter.go
```diff
@@ -37,6 +37,7 @@ type Registry interface {
 	GetHistogram(metricName string, labels prometheus.Labels, help string, mapping *mapper.MetricMapping, metricsCount *prometheus.GaugeVec) (prometheus.Observer, error)
 	GetSummary(metricName string, labels prometheus.Labels, help string, mapping *mapper.MetricMapping, metricsCount *prometheus.GaugeVec) (prometheus.Observer, error)
 	RemoveStaleMetrics()
+	ClearMetrics() int
```
pedro-stanaka, line 53: Adding `ClearMetrics() int` to the exported `exporter.Registry` interface is source-incompatible for downstream users that provide custom registry implementations or mocks. Because this repository is also consumed as a library, this change can cause immediate compile failures on upgrade even when runtime behavior is otherwise correct.

**Suggestion:** Consider keeping the exported `Registry` contract unchanged and introducing an internal optional capability interface (for example, `type clearableRegistry interface { ClearMetrics() int }`) behind a type assertion where clearing is needed. This preserves compatibility for existing library consumers.

README.md
```diff
@@ -113,8 +113,8 @@ NOTE: Version 0.7.0 switched to the [kingpin](https://github.com/alecthomas/king
 
 ## Lifecycle API
 
-The `statsd_exporter` has an optional lifecycle API (disabled by default) that can be used to reload or quit the exporter 
-by sending a `PUT` or `POST` request to the `/-/reload` or `/-/quit` endpoints.
+The `statsd_exporter` has an optional lifecycle API (disabled by default) that can be used to reload, quit, or clear the exporter
```
pedro-stanaka:
```suggestion
The `statsd_exporter` has an optional lifecycle API (disabled by default) that can be used to reload, quit, or clear dynamically registered StatsD metric series
```

pedro-stanaka (changes requested):
Need some changes to fix potential bugs and breakage of interface for library users.

README.md
yuminn-k: Applied, thanks.

pkg/exporter/exporter.go
yuminn-k: Done. I restored the exported `Registry` interface so downstream custom implementations and mocks do not need to add `ClearMetrics`.

Clearing is now handled through an internal optional `clearableRegistry` capability interface, with `ErrClearMetricsUnsupported` returned if the configured registry does not support it. I also added a test registry without `ClearMetrics` to cover this compatibility case.

pkg/exporter/exporter.go
yuminn-k: Thanks, good catch. I updated `ClearMetrics` to take a `context.Context` and return `(int, error)` so callers do not block indefinitely when the exporter listener is not running.

It now tracks the `Listen` lifecycle, returns `ErrExporterNotRunning` before/after the event loop, and still serializes successful clear requests through the exporter loop. The HTTP handler now returns `503` if clearing cannot be performed.

pkg/exporter/exporter.go
```diff
@@ -73,6 +115,51 @@ func (b *Exporter) Listen(e <-chan event.Events) {
...
+	}
+}
+
+func (b *Exporter) listenState() (<-chan struct{}, bool) {
+	b.listenStateMtx.RLock()
+	defer b.listenStateMtx.RUnlock()
+	return b.listenDone, b.listening
+}
+
+func (b *Exporter) clearRegistryMetrics() clearMetricsResult {
```
pedro-stanaka, line 155: I wonder if we should instrument this, for large metric sets, we could block for a while and cause the whole exporter to get "stuck" here. Maybe lets add at least a doc block saying that this function may cause "hangs" when working with larger sets.

main.go
```diff
@@ -205,6 +206,25 @@ func reloadConfig(fileName string, mapper *mapper.MetricMapper, logger *slog.Log
 	}
 }
 
+type metricsClearer interface {
+	ClearMetrics(context.Context) (int, error)
+}
+
+func clearMetricsHandler(clearer metricsClearer, logger *slog.Logger) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if r.Method == http.MethodPut || r.Method == http.MethodPost {
```
pedro-stanaka, line 215: Lets return 405 Method Not Allowed when the method is not POST|PUT.

Allow: PUT, POST, and test both supported methods and at least one unsupported method.

pedro-stanaka (commented):
Just a few more comments and it should be good to go, also sign the DCO please in your commits.

PR #715, opened 2026-06-25
feat: add current metrics gauge

### LINKED ISSUE #565: metricsCount metric remain the same after metric been deleted by RemoveStaleMetrics
I was using https://github.com/prometheus/statsd_exporter/blob/58769c7b4d128e7ceae1cf8d893260ed5b5afa4d/main.go#L167-L173 to track the metric number of statsd exporter, but found that this metric never drops, even if I set the ttl parameter in config file.

Then I noticed, across all the reference of the metrics, these's no place to decrease the gauge.

even in the https://github.com/prometheus/statsd_exporter/blob/58769c7b4d128e7ceae1cf8d893260ed5b5afa4d/pkg/registry/registry.go#L381-L396

So I think if we add a decrease in this function, the metric number would show the real number of metric number

yuminn-k:
## Summary
- add `statsd_exporter_metrics_current` to report the current number of active metric vectors by metric type
- keep `statsd_exporter_metrics_total` as the cumulative vector counter
- update the current gauge when metric vectors become active again or expire via TTL cleanup

Fixes #565

## Testing
- `go test ./pkg/registry ./pkg/exporter -run 'TestMetricsCurrentTracksActiveMetricVectors|TestTtlExpiration|TestHashLabelNames' -count=1`\n- `go test ./pkg/exporter -run TestMetricsCurrentTracksActiveMetricVectors -race -count=1`\n- `go test ./... -count=1`\n- `git diff --check`

### DIFF
--- a/main.go
+++ b/main.go
@@ -171,6 +171,13 @@ var (
 		},
 		[]string{"type"},
 	)
+	metricsCurrent = promauto.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "statsd_exporter_metrics_current",
+			Help: "The current number of metric vectors with active time series.",
+		},
+		[]string{"type"},
+	)
 )
 
 func serveHTTP(mux http.Handler, listenAddress string, logger *slog.Logger) {
@@ -326,7 +333,7 @@ func main() {
 		}
 	}
 
-	exporter := exporter.NewExporter(prometheus.DefaultRegisterer, thisMapper, logger, eventsActions, eventsUnmapped, errorEventStats, eventStats, conflictingEventStats, metricsCount)
+	exporter := exporter.NewExporter(prometheus.DefaultRegisterer, thisMapper, logger, eventsActions, eventsUnmapped, errorEventStats, eventStats, conflictingEventStats, metricsCount, metricsCurrent)
 
 	if *checkConfig {
 		logger.Info("Configuration check successful, exiting")
--- a/pkg/exporter/exporter_test.go
+++ b/pkg/exporter/exporter_test.go
@@ -140,6 +140,13 @@ var (
 		},
 		[]string{"type"},
 	)
+	metricsCurrent = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "statsd_exporter_metrics_current",
+			Help: "The current number of metric vectors with active time series.",
+		},
+		[]string{"type"},
+	)
 )
 
 // TestNegativeCounter validates when we send a negative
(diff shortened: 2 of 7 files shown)


main.go
```diff
@@ -171,6 +171,13 @@ var (
 		},
 		[]string{"type"},
 	)
+	metricsCurrent = promauto.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "statsd_exporter_metrics_current",
+			Help: "The current number of metric vectors with active time series.",
+		},
+		[]string{"type"},
+	)
```
pedro-stanaka, line 180: Worth tightening this metric name and help text before release. statsd_exporter_metrics_current can be read as the current number of metrics, active time series, or current metric values, but the implementation increments only when a metric vector’s refcount transitions from zero to non-zero. That means one metric name plus label-name set contributes one, regardless of how many label-value series are active. Since scrape-facing metric names become a long-lived operator API, this ambiguity will be hard to fix later.

Suggestion: Consider renaming the metric and updating the help text, tests, and changelog entry together. For example:


```suggestion
	metricsCurrent = promauto.NewGaugeVec(
		prometheus.GaugeOpts{
			Name: "statsd_exporter_active_metric_vectors",
			Help: "Current number of metric vectors that have at least one active time series.",
		},
		[]string{"type"},
	)
```

pkg/registry/registry.go
pedro-stanaka, line 1: Consider keeping the exported Registry API source-compatible while adding this telemetry. The PR changes GetCounter, GetGauge, GetHistogram, GetSummary, and RemoveStaleMetrics to require a metricsCurrent gauge. That will break downstream users of pkg/registry, and it also makes callers pass exporter self-metric plumbing through every lookup and cleanup call. Since Registry already owns Metrics, Vectors, and RefCount, it is a natural place to hold an optional current-gauge field internally.

We should keep the existing Get* and RemoveStaleMetrics signatures stable. Add the current gauge to registry state instead, for example via a new opt-in constructor or setter such as NewRegistryWithMetricsCurrent or SetMetricsCurrent, and have Store/RemoveStaleMetrics use that optional field internally. Existing direct callers would continue to compile and would simply get no current-gauge updates unless they opt in.

pkg/exporter/exporter_test.go
```diff
@@ -1168,6 +1175,62 @@ mappings:
 	}
 }
 
+func TestMetricsCurrentTracksActiveMetricVectors(t *testing.T) {
```
pedro-stanaka, line 1178: We should  extend the test after the inactive assertion to re-register the same metric vector and assert that the current gauge returns to 1.

pedro-stanaka (changes requested):
The changes are too broad and introduce breaking changes to the users of project as a library. Out of curiosity, how do you use the statsd-exporter in your work/day-to-day?

PR #716, opened 2026-06-30
pkg/line: support DogStatsD v1.3 metric timestamp field (|T<unix_ts>)

### LINKED ISSUE #680: DogStatsD v1.3 support an optional Unix timestamp field.

icohangar-ops:
Fixes #680.

Closes #684  

This finishes and supersedes the stalled #684 (by @alliasgher), who is credited via a `Co-authored-by` trailer on the commit. @pedro-stanaka invited a take-over in the issue thread.

## What

DogStatsD v1.3 adds an optional **metric timestamp** field:

```
<METRIC_NAME>:<VALUE>|<TYPE>|#<tags>|T<unix_timestamp>
```

The timestamp is a Unix UTC epoch in seconds, prefixed by `T` (e.g. `page.views:15|c|#env:dev|T1656581400`). The line parser previously rejected or mis-parsed datagrams carrying this field.

## How

- In `pkg/line/line.go`, the `|T<ts>` token is handled alongside the other optional pipe-delimited fields (`@` sampling, `#` tags, `c:` container ID). It is validated as a base-10 `int64` (Unix epoch) for protocol correctness via a new `case 'T'`.
- **Parse-and-drop**: per the maintainer guidance in #680, the timestamp is validated and then dropped. statsd_exporter exposes metrics over a pull-based Prometheus endpoint and does not forward per-sample source timestamps, so honoring it would not be meaningful. This keeps the change minimal and avoids touching `pkg/event`.
- Malformed/empty/non-numeric timestamps are rejected gracefully with a new `malformed_timestamp` sample error, consistent with how other optional fields (e.g. container ID) report errors.
- The component-count guard was tightened from `> 5` to `> 6` to admit a full v1.3 line (`value|type|@sampling|#tags|c:container|T<ts>` = 6 components) while still rejecting genuinely over-long lines. The existing "too many components" test still passes.

## Tests

Added table-driven cases in `pkg/line/line_test.go`: timestamped counter, timestamped gauge, timestamp without tags, timestamp combined with sampling + tags + container ID, and two malformed-timestamp cases (empty value and non-numeric).

```
go build ./...        # ok
go test ./...         # all packages pass
golangci-lint run     # 0 issues (v2.11.4, via make)
gofmt / go vet        # clean
```

## DCO

Commit is signed off (`git commit -s`). This repo uses DCO only (no CLA bot).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+* [FEATURE] Support the optional DogStatsD v1.3 metric timestamp field (`|T<unix_timestamp>`) ([#680](https://github.com/prometheus/statsd_exporter/issues/680))
+
 ## 0.30.0 / 2026-05-28
 * [CHANGE] Remove the Dockerfile `HEALTHCHECK` from published container images ([#671](https://github.com/prometheus/statsd_exporter/pull/671))
 * [ENHANCEMENT] Add a distroless container image variant ([#703](https://github.com/prometheus/statsd_exporter/pull/703))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Unreleased
 
-* [FEATURE] Support the optional DogStatsD v1.3 metric timestamp field (`|T<unix_timestamp>`) ([#680](https://github.com/prometheus/statsd_exporter/issues/680))
+* [FEATURE] Support the optional DogStatsD v1.3 metric timestamp field (`|T<unix_timestamp>`) ([#716](https://github.com/prometheus/statsd_exporter/pull/716))
 
 ## 0.30.0 / 2026-05-28
 * [CHANGE] Remove the Dockerfile `HEALTHCHECK` from published container images ([#671](https://github.com/prometheus/statsd_exporter/pull/671))
--- a/pkg/line/line.go
+++ b/pkg/line/line.go
@@ -269,7 +269,7 @@ samples:
 	for _, sample := range samples {
 		samplesReceived.Inc()
 		components := strings.Split(sample, "|")
-		if len(components) < 2 || len(components) > 5 {
+		if len(components) < 2 || len(components) > 6 {
 			sampleErrors.WithLabelValues("malformed_component").Inc()
 			logger.Debug("bad component", "line", line)
 			continue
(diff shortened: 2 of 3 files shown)


pkg/line/line.go
```diff
@@ -329,6 +329,22 @@ samples:
...
+					// The value is a Unix UTC epoch in seconds prefixed by 'T'.
+					// statsd_exporter exposes metrics over a pull-based Prometheus
+					// endpoint and does not forward per-sample source timestamps, so
+					// we validate the field for protocol correctness and then drop it.
+					if len(component) < 2 {
+						logger.Debug("Malformed timestamp section", "component", component, "line", line)
+						sampleErrors.WithLabelValues("malformed_timestamp").Inc()
+						continue samples
+					}
+					if _, err := strconv.ParseInt(component[1:], 10, 64); err != nil {
```
pedro-stanaka: Please check feedback from previous PR.

pedro-stanaka (changes requested):

pkg/line/line.go
pedro-stanaka: I had a review ready on the other PR, but forgot to submit. This version looks good, just want to drop this Parse here, it will waste CPU cycles.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## Unreleased
+
+* [FEATURE] Support the optional DogStatsD v1.3 metric timestamp field (`|T<unix_timestamp>`) ([#680](https://github.com/prometheus/statsd_exporter/issues/680))
```
pedro-stanaka: This will be confusing, lets link the current PR.

pkg/line/line.go
icohangar-ops: Done in ff7ed49 — dropped the `strconv.ParseInt` entirely. `case 'T':` is now a pure no-op drop with a comment explaining why: since statsd_exporter is pull-based and discards the per-sample timestamp anyway, parsing/validating it only wastes CPU on the hot path. I also removed the now-unused empty-value guard and the `malformed_timestamp` sample-error label, and updated the tests so an invalid or empty timestamp is simply ignored while the metric still parses.

CHANGELOG.md
icohangar-ops: Done in ff7ed49 — relinked the CHANGELOG entry to the PR ([#716](https://github.com/prometheus/statsd_exporter/pull/716)) instead of the issue.

pkg/line/line.go
icohangar-ops: Addressed — see the no-op drop in ff7ed49. The parse is gone from the hot path.

icohangar-ops:
Thanks for the review @pedro-stanaka! Pushed ff7ed49:

- `case 'T':` is now a pure no-op drop — removed the `strconv.ParseInt` validation (and the now-unused `malformed_timestamp` error + empty guard) so we don't burn CPU parsing a timestamp we discard anyway.
- Tests updated: invalid/empty timestamp values are now just dropped while the metric parses normally.
- CHANGELOG entry relinked to #716 instead of the issue.

`go build`, `go test ./...`, `go vet`, `gofmt -l`, and `golangci-lint run` are all green locally. Commit is DCO-signed.

pedro-stanaka (approved):

icohangar-ops:
Thanks Pedro



Sam Desigan

PR #735, opened 2026-07-31
Support $0 template reference in glob mappings

### LINKED ISSUE #444: Allow templating with `$0` in glob mode
Split from #438: in the regex match type, it is possible to use `$0` to reference the original metric name. Since the glob type _used_ to be translated to regular expressions it was possible there as well, but we lost this feature with the transition to a finite state machine. We should bring it back.

pujitha24:
Motivation:
Split from #438, issue #444: with regex match_type, `$0` in a
mapping's `name` or a label value expands to the original,
unmodified statsd metric name via Go's regexp.Expand semantics.
Glob mappings used to be translated to regular expressions
internally, so `$0` also worked there, but that stopped working
when glob matching was rewritten to use a finite state machine
(FSM), since the FSM only tracks captured `*` segments, not the
full input metric name. Before this change, using `$0` with
`match_type: glob` silently expanded to an empty string (the
same fallback used for any out-of-range `$N` reference) rather
than erroring, so the loss was easy to miss.

Approach:
- pkg/mapper/fsm/formatter.go: TemplateFormatter.NewTemplateFormatter
  now accepts index 0 (previously rejected as `idx < 1`) and stores
  it as a sentinel (-1) alongside the existing capture indexes.
  TemplateFormatter.Format now takes the original metric name as an
  additional argument and substitutes it whenever the sentinel is
  seen, leaving existing $1..$N handling unchanged.
- pkg/mapper/mapper.go: pass the original statsdMetric string through
  to nameFormatter.Format and each label formatter's Format call.
- Updated README.md to document that `$0` now works for both `glob`
  and `regex` match types, and to note (matching existing $N
  behavior) that `${0}` bracing is required when immediately
  followed by other word characters or another `$`-reference.

Validation:
- go build ./... and go test ./... pass.
- Added a table-driven case to pkg/mapper/mapper_test.go
  (TestMetricMapperYAML) covering `$0` in both the metric `name`
  and a label value for glob mappings; verified this new case fails
  with the pre-fix code (empty label value) and passes with the fix.
- This is a feature restoration matching existing regex-mode `$0`
  behavior, not a crash/data-loss fix: before the fix, glob-mode
  `$0` silently resolved to an empty string, it did not error or
  panic.

Fixes #444

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 
 * [FEATURE] Support the optional DogStatsD v1.3 metric timestamp field (`|T<unix_timestamp>`) ([#716](https://github.com/prometheus/statsd_exporter/pull/716))
+* [FEATURE] Support `$0` template reference (the original, unmodified statsd metric name) in `glob` mappings, matching existing `regex` mapping behavior ([#735](https://github.com/prometheus/statsd_exporter/pull/735))
 
 ## 0.30.0 / 2026-05-28
 * [CHANGE] Remove the Dockerfile `HEALTHCHECK` from published container images ([#671](https://github.com/prometheus/statsd_exporter/pull/671))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 ## Unreleased
 
 * [FEATURE] Support the optional DogStatsD v1.3 metric timestamp field (`|T<unix_timestamp>`) ([#716](https://github.com/prometheus/statsd_exporter/pull/716))
-* [FEATURE] Support `$0` template reference (the original, unmodified statsd metric name) in `glob` mappings, matching existing `regex` mapping behavior ([#735](https://github.com/prometheus/statsd_exporter/pull/735))
+* [FEATURE] Support `$0` template reference in `glob` mappings, expanding to the complete StatsD metric name (matching regex mappings, where `$0` expands to the complete regex match) ([#735](https://github.com/prometheus/statsd_exporter/pull/735))
+* [CHANGE][library] `fsm.TemplateFormatter.Format` now takes the original metric name as an additional first argument: `Format(metricName string, captures []string)` instead of `Format(captures []string)` ([#735](https://github.com/prometheus/statsd_exporter/pull/735))
 
 ## 0.30.0 / 2026-05-28
 * [CHANGE] Remove the Dockerfile `HEALTHCHECK` from published container images ([#671](https://github.com/prometheus/statsd_exporter/pull/671))
--- a/README.md
+++ b/README.md
@@ -277,7 +277,7 @@ But the metric will also have the label `statsd_metric_name` with the value `my.
 
 Note: If you use the `match` like the example (i.e. `.+`), be aware that it will be a "catch-all" block. So it should come at the very end of the mapping list.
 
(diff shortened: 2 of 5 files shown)


pujitha24:
This has been rebased and green for a while now — happy to make any changes that would help move review along.

pedro-stanaka:
Please add tests for: 

- $0 in the metric name, not only labels.
- Combined $0 and $1.
- Out-of-range $N.
- $0 with a literal/no-wildcard glob.
- ${0}_suffix bracing behavior.

And add a changelog.

pujitha24:
Added test cases for each of these: `$0` in the metric name (not just labels), `$0` combined with `$1`, an out-of-range `$N` alongside `$0`, `$0` with a literal/no-wildcard glob, and the `$0_suffix` vs `${0}_suffix` bracing behavior. Also added a changelog entry under Unreleased. All existing and new tests pass.

README.md
```diff
@@ -258,7 +260,7 @@ mappings:
 
 #### Special match groups
 
-When using regex, the match group `0` is the full match and can be used to attach labels to the metric.
+The `$0` reference expands to the original, unmodified statsd metric name and can be used with both `glob` and `regex` matching to attach labels to the metric.
```
pedro-stanaka: For `glob`, `$0` is the complete input metric name, but for `regex` it remains regex match group 0 and may be only a substring because regex mappings are not implicitly anchored (for example, matching `foo` against `foo.bar` gives `$0 == "foo"`). Could we preserve that distinction here? Suggested wording: `With glob, $0 expands to the complete StatsD metric name. With regex, $0 expands to the complete regex match.` The `${0}` bracing guidance below should also be scoped specifically to glob templates.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Unreleased
 
 * [FEATURE] Support the optional DogStatsD v1.3 metric timestamp field (`|T<unix_timestamp>`) ([#716](https://github.com/prometheus/statsd_exporter/pull/716))
+* [FEATURE] Support `$0` template reference (the original, unmodified statsd metric name) in `glob` mappings, matching existing `regex` mapping behavior ([#735](https://github.com/prometheus/statsd_exporter/pull/735))
```
pedro-stanaka: Could we also call out the breaking library change here? `TemplateFormatter.Format` changes from `Format(captures []string)` to `Format(metricName string, captures []string)`, so downstream users importing `pkg/mapper/fsm` will fail to compile. This should have a `[CHANGE][library]` entry (or preserve the old method and add a new `FormatWithMetricName` method). Also, `matching existing regex mapping behavior` has the same full-input-vs-full-match distinction noted in the README comment.

pedro-stanaka (commented):
Thanks for adding the expanded tests and changelog entry. I still see two documentation/API-contract details to address inline.

README.md
```diff
@@ -275,7 +277,7 @@ But the metric will also have the label `statsd_metric_name` with the value `my.
 
 Note: If you use the `match` like the example (i.e. `.+`), be aware that it will be a "catch-all" block. So it should come at the very end of the mapping list.
 
-The same is not achievable with glob matching, for more details check [this issue](https://github.com/prometheus/statsd_exporter/issues/444).
+Note: as with other `$n` references (for both `glob` and `regex` templates), use the `${0}` form (rather than `$0`) when it is immediately followed by other word characters or another `$`-reference, e.g. `${0}_total`, otherwise the reference will not be recognized and will resolve to an empty string.
```
pedro-stanaka: nit: for regex $0$1 works as well (no braces). Wanna fix that? Then I can merge it.

pedro-stanaka (approved):

pedro-stanaka (approved):

PR #736, opened 2026-08-01
feat: support $0 in glob mapping templates

### LINKED ISSUE #444: Allow templating with `$0` in glob mode

SaiPisey2:
Closes #444

In the regex match type `$0` references the whole metric name. Glob mappings supported this too, back when they were translated into regular expressions, but the move to a finite state machine dropped it.

`NewTemplateFormatter` rejects any reference below `$1` and silently replaces it with an empty string:

```go
if err != nil || idx > captureCount || idx < 1 {
    valueFormatter = strings.ReplaceAll(valueFormatter, match[0], "")
```

So for `test.dispatcher.FooProcessor.send` with `name: "prefix_$0"`, the two match types disagree:

| match type | resulting name |
| --- | --- |
| regex | `prefix_test.dispatcher.FooProcessor.send` |
| glob | `prefix_` |

## Change

The metric name is passed to the formatters as the first element of the capture slice, and the reference now indexes into it directly. `$0` is the whole metric name and `$1` onwards keep pointing at exactly the same captures as before. After the change both match types produce `prefix_test.dispatcher.FooProcessor.send`.

The change is contained: `NewTemplateFormatter` only has two callers and both are on the glob path. The regex path uses `regexp.ExpandString`, which already handles `$0` natively and is untouched.

Out-of-range references keep their existing behaviour — `$2` on a single-capture match still renders as an empty string.

Worth noting for review: the mapping cache is keyed on the full metric name (`formatKey(statsdMetric, statsdMetricType)`), so results containing a per-metric `$0` cache correctly rather than being shared across metrics that hit the same mapping.

## Testing

Added a scenario to `TestMetricMapperYAML` covering `$0` on its own, `$0` combined with `$1`, and an out-of-range reference. Verified locally:

- `go build ./...`
- `go vet ./...`
- `gofmt -l` clean
- `go test ./...` — full suite passes, no changes needed to existing expectations

## Docs

The README documented this as a limitation and linked to #444, so that note is replaced with a short glob example.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 
 * [FEATURE] Support the optional DogStatsD v1.3 metric timestamp field (`|T<unix_timestamp>`) ([#716](https://github.com/prometheus/statsd_exporter/pull/716))
+* [FEATURE] Support `$0` to reference the whole metric name in glob mapping templates ([#736](https://github.com/prometheus/statsd_exporter/pull/736))
 
 ## 0.30.0 / 2026-05-28
 * [CHANGE] Remove the Dockerfile `HEALTHCHECK` from published container images ([#671](https://github.com/prometheus/statsd_exporter/pull/671))
--- a/README.md
+++ b/README.md
@@ -258,7 +258,7 @@ mappings:
 
 #### Special match groups
 
-When using regex, the match group `0` is the full match and can be used to attach labels to the metric.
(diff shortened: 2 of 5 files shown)


pedro-stanaka:
Closing in favor of #735

PR #746, opened 2026-08-17
fix(fsm): support prefixed glob matches via wildcard traversal

### LINKED ISSUE #481: Support prefixed glob matches

AruneshDwivedi:
## Problem

When using glob match patterns with wildcards in positions other than
the end (e.g. 'prefix.*.metric'), the FSM does not traverse deeper
into the automaton because * is treated as a terminal capture state.
This means patterns like 'app.*.metric' cannot match metrics with
additional path segments like 'app.foo.bar.metric'.

Fixes #481

## Solution

Add a Prefix flag to mappingState. When a * wildcard is added via
AddState, mark the state as Prefix=true. In GetMapping, when no
exact or wildcard match is found at the current position, check if
the current state is a Prefix state with deeper transitions - if so,
continue traversing instead of breaking.

### DIFF
--- a/pkg/mapper/fsm/fsm.go
+++ b/pkg/mapper/fsm/fsm.go
@@ -26,6 +26,10 @@ type mappingState struct {
 	// result* members are nil unless there's a metric ends with this state
 	Result         interface{}
 	ResultPriority int
+	// Prefix denotes a wildcard state that acts as a prefix matcher.
+	// When true, the FSM may traverse through this state into
+	// deeper transitions even if no result is anchored at this state.
+	Prefix bool
 }
 
 type fsmBacktrackStackCursor struct {
@@ -105,6 +109,10 @@ func (f *FSM) AddState(match, matchMetricType string, maxPossibleTransitions int
 			}
 			if field == "*" {
 				captureCount++
+				// Mark as prefix state so GetMapping knows to traverse deeper
+				// This allows patterns like "prefix.*.metric" to match any
+				// number of segments after the prefix
+				state.Prefix = true
 			}
 
 			// goto next state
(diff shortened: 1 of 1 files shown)


pedro-stanaka (changes requested):
Thanks for taking this on. I’m blocking this version because it changes the meaning of the existing component wildcard and can leave `state == nil`, causing ordinary non-matches to panic in `GetMapping`. The existing mapper test suite reproduces this regression, when creating automated PRs or PRs with help of tools, please verify locally by running tests.

Issue #481 asks for a wildcard **within one dot-separated component**: `prefixed_glob.prefix_*.*` should match `prefixed_glob.prefix_myprefix.mymetric`, with `$1 = myprefix` and `$2 = mymetric`. It is not asking for `*` to consume multiple components.

I think the correct implementation shape is to model `prefix_*` as a new component-level transition alongside exact and bare-`*` transitions:

   - Match one component using `strings.HasPrefix`.
   - Capture only the suffix after the literal prefix.
   - Preserve the existing component count and `min/maxRemainingLength` behavior.
   - Include exact, matching-prefix, and bare-`*` edges in backtracking.
   - In unordered mode, prefer exact → longest matching prefix → bare `*`; in ordered mode, preserve declaration priority.
   - Keep validation narrow to one trailing `*` rather than accepting arbitrary embedded or repeated wildcards.

pedro-stanaka:
Thanks for your interest in working on this. Before submitting another implementation, please pause and engage with us on issue #481 so we can agree on the intended semantics and design first. This PR does not address the behavior requested by the issue, introduces a regression in the existing test suite, and suggests the relevant code paths were not fully validated.

We have (very) limited maintainer review capacity, so speculative PRs without prior design discussion create substantial review overhead. Please continue the conversation on the issue, demonstrate an understanding of the existing FSM, and wait for maintainer alignment before opening a replacement PR. 

Tool-assisted contributions are welcome, but contributors remain responsible for understanding, testing, and supporting the changes they submit. For now, I’m closing this PR.

PR #747, opened 2026-08-20
fix: clear registry on config reload to pick up new bucket settings

### LINKED ISSUE #471: bug: Configuration reload does not affect bucket configuration
## Bug description

**What is happening?**

1. I start the exporter with a mapping configuration (see example config at the end).
2. I send some samples of the metric constantly (before metric had chance to churn)
3. I change mapping configuration and update the bucked layout for my metrics
4. I trigger a reload via sighup/web lifecycle
5. After sending new samples, old bucket configuration is still present and not new, updated one.

This is happening because the metric is never deleted after the reload from the `Registry`.

**What you expected to happen?**

On step `5.` I would expect the new bucket configuration to take place.

### Notes

Example config:

```
defaults:
    ttl: 3m
    match_type: regex
    observer_type: histogram
    histogram_options:
        buckets: [0.1, 0.25, 150, 500, 900]
mappings:
    - match: \bexample_metric\b
      observer_type: histogram
      histogram_options:
        buckets: [10, 40, 50, 80, 100, 120, 150, 200, 250]
      name: example_metric
      labels:
        statsd_exporter_config_mapping: "true"
```

AruneshDwivedi:
Fixes #471

When the mapping configuration is reloaded via SIGHUP or lifecycle API, the MetricMapper is reinitialized but the Registry retains cached histogram/summary vectors with the old bucket or quantile settings. This causes the exporter to continue using stale configuration after reload.

Changes:
- Add Reset() method to the Registry interface that clears all cached metrics
- Call registry.Reset() after successful config reload in both SIGHUP and lifecycle API paths
- Add unit tests for the Reset() method

Root cause: The Registry caches prometheus.HistogramVec and prometheus.SummaryVec objects. These are created once with the bucket/quantile configuration at the time of first use. When the mapping config is reloaded with different bucket settings, the old vectors are still returned for subsequent metric lookups.

### DIFF
--- a/main.go
+++ b/main.go
@@ -178,7 +178,7 @@ func serveHTTP(mux http.Handler, listenAddress string, logger *slog.Logger) {
 	os.Exit(1)
 }
 
-func sighupConfigReloader(fileName string, mapper *mapper.MetricMapper, logger *slog.Logger) {
+func sighupConfigReloader(fileName string, mapper *mapper.MetricMapper, registry exporter.Registry, logger *slog.Logger) {
 	signals := make(chan os.Signal, 1)
 	signal.Notify(signals, syscall.SIGHUP)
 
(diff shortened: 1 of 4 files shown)


pedro-stanaka:
I dont think this will be needed anymore since:

* I believe for the short-future we should make native histograms the default
* People can use the new reset web lifecycle for this once it merges #714

PR #751, opened 2026-09-03
fix: elevate malformed tag log level from Debug to Warn

### LINKED ISSUE #552: Malformed name tag - Log level should be higher
The following log is currently set at debug level: https://github.com/prometheus/statsd_exporter/blob/58769c7b4d128e7ceae1cf8d893260ed5b5afa4d/pkg/line/line.go#L124 

This log is very useful for debugging missing tags in metrics. However to get it we have to set statsd-exporter at debug level logging, which also brings in the "incoming line" log which logs way too much causing issues for us. 

Can we increase this log to at least info level? Debug doesn't make sense to me for such important info

AruneshDwivedi:
Fixes #552

Malformed tag errors were only logged at Debug level, requiring users to enable
full debug logging to see important parsing failures.

Changed logger.Debug to logger.Warn for malformed tag errors in parseTag() so
these errors are visible in normal operation without needing debug-level logging.

### DIFF
--- a/pkg/line/line.go
+++ b/pkg/line/line.go
@@ -110,7 +110,7 @@ func parseTag(component, tag string, separator rune, labels map[string]string, t
 			if len(k) == 0 || len(v) == 0 {
 				// Empty key or value is an error
 				tagErrors.Inc()
-				logger.Debug("Malformed name tag", "k", k, "v", v, "component", component)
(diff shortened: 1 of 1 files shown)


pedro-stanaka:
If some part of the client of the statsd_exporter starts sending malformed packages it could hinder the performance for the whole application. It is better to troubleshoot this by flipping the flag when needed than changing the level like this.

PR #752, opened 2026-09-03
feat: add ClearRegistry endpoint to clear all metrics immediately

### LINKED ISSUE #637: lifecycle web endpoint for clearing the registry

AruneshDwivedi:
Fixes #637

Adds a ClearRegistry method to the Registry interface that immediately removes
all registered metrics. This allows users to clear the metric registry without
quitting the exporter.

Use case: When working with large clusters of statsd_exporter where the number
of instances changes or order changes, metrics distribution keys change due to
hashing on statsrelay. Users need a way to immediately forget accumulated
metrics and start collecting in a new distribution.

This is different from /-/reload (metrics persist) and /-/quit (disruptive).

### DIFF
--- a/pkg/exporter/exporter.go
+++ b/pkg/exporter/exporter.go
@@ -37,6 +37,7 @@ type Registry interface {
 	GetHistogram(metricName string, labels prometheus.Labels, help string, mapping *mapper.MetricMapping, metricsCount *prometheus.GaugeVec) (prometheus.Observer, error)
 	GetSummary(metricName string, labels prometheus.Labels, help string, mapping *mapper.MetricMapping, metricsCount *prometheus.GaugeVec) (prometheus.Observer, error)
 	RemoveStaleMetrics()
+	ClearRegistry()
 }
 
 type Exporter struct {
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -395,6 +395,25 @@ func (r *Registry) RemoveStaleMetrics() {
 	}
 }
 
+func (r *Registry) ClearRegistry() {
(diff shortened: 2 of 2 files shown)


pedro-stanaka:
We already have #747 which is working on that, I would rather work with that first for now.

